PR #8, opened 2017-07-25
Flask Instrumentation through WSGI middleware.

pglombardo:
WSGI is the equivalent to rack in the ruby world.  Here we add WSGI middleware that can be used to instrument many frameworks that use WSGI (including Flask).

Remaining issues:
- [x] App name is not properly reported
- [x] When runtime only, we should report process name as name (currently `sensory.pyc`)
- [x] Context headers are not currently returned.


pglombardo:
instana python flask 2017-07-27 at 14 38 22
instana python flask 2017-07-27 at 14 39 18

pglombardo:
instana go to python flask trace 2017-07-27 at 15 21 13

PR #10, opened 2017-07-27
Backport Django support to older version 1.9

pglombardo:
This adds support for Django 1.9 tracing.


pglombardo:
Better Django app name extraction:
screen shot 2017-07-28 at 12 02 51

A go app calling a Django 1.9 app:
instana django 1 9 from go app 2017-07-28 at 12 02 31

PR #18, opened 2017-09-15
Fixes a bug with connecting to the agent

elgris:
Basically, it fixes 2 pieces:

1. Hanging shell call in Docker container (as discovered, we need a `shell=True` there). Also the command itself was taken from golang sensor
2. One small bug when value must be updated before the condition

### DIFF
--- a/instana/fsm.py
+++ b/instana/fsm.py
@@ -66,7 +66,7 @@ class Fsm(object):
         else:
             host = self.get_default_gateway()
             if host:
-                self.check_host(host)
+                h = self.check_host(host)
                 if h == a.AGENT_HEADER:
                     self.agent.set_host(host)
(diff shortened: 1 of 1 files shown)


pglombardo:
A debug log surfaced in python 3.6.2.  I'll fix in a separate PR before release:
https://travis-ci.org/instana/python-sensor/jobs/275808812

elgris:
thanks @pglombardo !
For me changing a line `fsm.py:86` from
```
proc.stdout.read()
```
to
```
str(proc.stdout.read())
```
fixed Python3 tests

pglombardo:
That might be the more appropriate place.  I did the same just in the actual debug line in `check_host`  with `str()`.

PR #20, opened 2017-09-17
Implement advanced announce for containers

pglombardo:
This PR fixes the many pid == 1 issue when running inside of containers similar to what we do in other languages.

Remaining:
- [x] Don't attempt advanced announce when no `/proc` filesystem (Windows, OSX)


pglombardo:
screen shot 2017-09-17 at 18 03 12

PR #29, opened 2017-10-10
Propagators should support all forms of Instana Headers

pglombardo:
Depending on where `Inject` and `Extract` are being called from, the HTTP headers can vary in capitalization and prefix.

For example, in Flask and Django, in the middleware, HTTP headers are presented as all upper case, underscores and prefixed with `HTTP_` (Ruby also does this in Rack):

* `HTTP_X_INSTANA_T`

Yet in making requests via the `requests` package, HTTP headers are in the on-the-wire format:

* `X-Instana-T`

Our propagators should support all variations:

* Capitalization
* Underscores versus dashes
* Prefixed or not (only support `HTTP_` for now)

### DIFF
--- a/instana/http_propagator.py
+++ b/instana/http_propagator.py
@@ -22,8 +22,10 @@ class HTTPPropagator():
 
     HEADER_KEY_T = 'X-Instana-T'
     HEADER_KEY_S = 'X-Instana-S'
+    HEADER_KEY_L = 'X-Instana-L'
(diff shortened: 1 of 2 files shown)


pglombardo:
With this PR, we support `Extract` as case insensitive search for `X-Instana-[TS]` and the alternate form `HTTP_X_INSTANA_[TS]`.

For Inject, we _only_ support inserting `X-Instana-[TS]` as OpenTracing has no mechanism to signal otherwise.

PR #58, opened 2018-01-21
Your Idea worked.

hughbrien:
With respect to the Instrumentation Testing 

I am not sure how to "test" the "async" code, because there needs to "another thread" 

The idea of for including Tornado worked great.

### DIFF
--- a/instana/instrumentation/instrument_aiohttp.py
+++ b/instana/instrumentation/instrument_aiohttp.py
@@ -1,0 +1,55 @@
+#Attempting to work with both versions
+
+from __future__ import absolute_import
+import opentracing.ext.tags as ext
+import instana
+import opentracing
+import wrapt
+
+
+
+try:
+    import instrument_aiohttp
+
+    @wrapt.patch_function_wrapper('aiohttp', 'client.ClientSession.get')
+    def urlopen_with_instana(wrapped, instance, args, kwargs):
+
+        print("Instrumenting aiohttp")
+        context = instana.internal_tracer.current_context()
+
+        # If we're not tracing, just return
+        if context is None:
+            return wrapped(*args, **kwargs)
+
+        try:
+            span = instana.internal_tracer.start_span("aiohttp", child_of=context)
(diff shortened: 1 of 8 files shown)


pglombardo:
Hey @hughbrien - thanks for the PR!  To start to merge this stuff though, let's break this into 3 separate PRs and then we can work, test & validate each one individually.

A couple points:
* Delete vcs.xml file - it's not needed for this repository
* Delete this PR and make 1 each for aiomysql, aiohttp & tornado
* Add tests for each

instana/instrumentation/instrument_aiohttp.py
```diff
@@ -0,0 +1,55 @@
...
+from __future__ import absolute_import
+import opentracing.ext.tags as ext
+import instana
+import opentracing
+import wrapt
+
+
+
+try:
+    import instrument_aiohttp
```
pglombardo: This should be attempting to import the aiohttp package.  Here you might have mistakenly put the module name itself (instrument_aiohttp).  If aiohttp is available then we instrument, if not then pass.

instana/instrumentation/instrument_aiomysql.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import absolute_import
+import instana
+import wrapt
+import opentracing.ext.tags as ext
+
+try:
+    import instrument_aiomysql
```
pglombardo: Same comment as aiohttp.

instana/instrumentation/instrument_tornado.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import print_function
+
+import opentracing.ext.tags as ext
+import opentracing as ot
+import instana
+import opentracing
+import wrapt
+
+try:
+	import instrument_tornado
```
pglombardo: Same comment as aiohttp.

hughbrien:
Good idea. I will re-fork and do it in three parts.

PR #76, opened 2018-05-25
Adding GRPC support for HTTP headers

mag415:
I'm trying to use the [opentracing grpc library](https://github.com/opentracing-contrib/python-grpc) with the instana tracer.
However, the HTTP headers created by instana tracer are considered to be invalid metadata by the grpc lib since it contains upper case characters.

https://github.com/grpc/grpc/blob/master/src/core/lib/surface/validate_metadata.cc
validate_metadata: {"created":"@ ... ","description":"Illegal header key","file":"src/core/lib/surface/validate_metadata.cc","file_line":43,"offset":0,"raw_bytes":" ... 'X-Instana-T'"}

The fix here is to add an optional type of lower case http headers.

### DIFF
--- a/instana/http_propagator.py
+++ b/instana/http_propagator.py
@@ -26,6 +26,19 @@ class HTTPPropagator():
     ALT_HEADER_KEY_T = 'HTTP_X_INSTANA_T'
     ALT_HEADER_KEY_S = 'HTTP_X_INSTANA_S'
     ALT_HEADER_KEY_L = 'HTTP_X_INSTANA_L'
+    GRPC_SAFE_HEADER_KEY_T = 'x-instana-t'
+    GRPC_SAFE_HEADER_KEY_S = 'x-instana-s'
+    GRPC_SAFE_HEADER_KEY_L = 'x-instana-l'
+    
+    def __init__(self, use_grpc_safe_headers):
+        if(use_grpc_safe_headers):
+            self.header_key_t = self.GRPC_SAFE_HEADER_KEY_T
+            self.header_key_s = self.GRPC_SAFE_HEADER_KEY_S
+            self.header_key_l = self.GRPC_SAFE_HEADER_KEY_L
+        else:
+            self.header_key_t = self.HEADER_KEY_T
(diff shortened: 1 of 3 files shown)


pglombardo:
Hi @mag415 - thanks for the PR and pointing this out.  Honestly I'm undecided on how to best handle this.  If another special case comes up, do we add another if case?  Or do we instead split these variations out into separate propagators?

I'm going to give this some thought and post back soon.  Thanks!

pglombardo:
One downside I see with this, is that to enable this, it's set via environment variable which makes this take affect globally for the application.  But what do we do if we have a mix of clients - both HTTP and GRPC?  We are stuck using one form or another.

Additionally, since OpenTracing [doesn't define a specific GRPC format type](https://github.com/opentracing/opentracing-python/blob/master/opentracing/propagation.py#L55-L78), we can't easily spin of a dedicated propagator for this.

And finally, it crossed my mind that we could provide a default parameter (`use_grpc_safe_headers=false`) to the `inject` method but that is veering off from [the OpenTracing APIs](https://github.com/opentracing/opentracing-python/blob/master/opentracing/tracer.py#L83).

So still undecided but progressing :-)

pglombardo:
Hi @mag415 after much personal debating on this issue, it boils down to a limitation in OpenTracing that it does not yet provide support for this use case.

The core problem with this PR is that if you set the environment variable and enable GRPC headers, it breaks HTTP context passing (no tracing of distributed HTTP calls).

I do have an alternative though that may work for you in the interim:  If you subclass `InstanaTracer`, you can set the [HTTPPropagator](https://github.com/instana/python-sensor/blob/master/instana/http_propagator.py#L20) to one of your own with GRPC safe headers.  The InstanaTracer is declared [here](https://github.com/instana/python-sensor/blob/master/instana/tracer.py#L15-L25).

I was hoping for a better outcome for this use case but for the above reasons, we can't support this use case in this fashion.

pglombardo:
Also for the record, we're soliciting input from the OpenTracing community with this issue: https://github.com/opentracing/opentracing-python/issues/87

PR #81, opened 2018-06-21
WIP: Start cleaning up globals

nyanshak:
Start to move some functionality out of global / package level variables, enabling more flexibility in using the library. With this change, you can make multiple tracers with different service names, rather than service name being based on instana.appname (module variable).

I haven't gone through to verify / test it yet but kind of shows where I'd like to take it if / when I have more time to finish.

### DIFF
--- a/instana/__init__.py
+++ b/instana/__init__.py
@@ -55,9 +55,6 @@ global_sensor = Sensor(Options())
 #
 internal_tracer = InstanaTracer()
 
-# Set ourselves as the tracer.
-opentracing.tracer = internal_tracer
-
 # Optional application wide service name.
 # Can be configured via environment variable or via code:
 #
--- a/instana/meter.py
+++ b/instana/meter.py
@@ -149,18 +149,7 @@ class Meter(object):
 
     def collect_snapshot(self):
         try:
-            if instana.service_name:
-                appname = instana.service_name
-            elif "FLASK_APP" in os.environ:
-                appname = os.environ["FLASK_APP"]
-            elif "DJANGO_SETTINGS_MODULE" in os.environ:
-                appname = os.environ["DJANGO_SETTINGS_MODULE"].split('.')[0]
-            elif hasattr(sys, '__interactivehook__') or hasattr(sys, 'ps1'):
-                appname = "Interactive Console"
-            else:
-                appname = os.path.basename(sys.argv[0])
-
-            s = Snapshot(name=appname, version=platform.version(),
+            s = Snapshot(name=self.sensor.options.service_name, version=platform.version(),
                          f=platform.python_implementation(),
                          a=platform.architecture()[0],
                          djmw=self.djmw)
--- a/instana/options.py
+++ b/instana/options.py
@@ -1,11 +1,13 @@
 import logging
 import os
+import sys
 
 
 class Options(object):
     service = ''
     agent_host = ''
     agent_port = 0
+    service_name = 'myservice'
     log_level = logging.WARN
 
     def __init__(self, **kwds):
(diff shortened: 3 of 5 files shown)


pglombardo:
Hi @nyanshak - thanks for the PR.  These are good changes.  I was also starting this locally. Two things to note:

1. See [here for the Python service naming strategy](https://github.com/instana/python-sensor/blob/service-naming-improvements/Configuration.md#custom-service-names) pending documentation.

2. Since `Sensor` is the part that collects reports metrics for the entire process, there should be only one for the entire process and in-turn shared by each `Tracer` instance.

3. `Snapshot.appname` is not necessarily always the service name.  This specifically affects the naming of the python process entity in the dashboard.

With those requirements above, these changes can become a bit complex so let me know if any questions but thanks for taking a stab at this!

instana/__init__.py
```diff
@@ -55,9 +55,6 @@ def load(module):
 #
 internal_tracer = InstanaTracer()
 
-# Set ourselves as the tracer.
-opentracing.tracer = internal_tracer
-
```
pglombardo: I'm curious as to the reason for this removal.  This allows OpenTracing plugability.  So existing calls in an application to `opentracing.tracer.start_span` etc just automatically work when the instana package is used.

pglombardo: I know this is WIP so I don't mean to overload :-)

nyanshak: In other OT libraries, it's more common (and I would argue less-surprising) to not automatically modify another package's variables. For example, with https://github.com/lightstep/lightstep-tracer-python, to set the opentracing.tracer, they recommend doing this:

```
  opentracing.tracer = lightstep.Tracer(
    component_name='your_microservice_name',
    access_token='{your_access_token}')
```
Although, yes, it is setting opentracing.tracer, it's giving the user of the library the option to set it when they want and not doing it automatically.

In probably MOST cases, this works fine, but I'm trying to test tracers to multiple systems. One way to do this is to build a multi-tracer, which calls the start_span (or any other methods) on an array of tracers. It's not quite that simple, but that's the basic idea.

nyanshak:
Sorry, I didn't have the full docs to know what these should be. :)

My use case is a bit abnormal so I would understand if you don't think it's worth pursuing. What I'm trying to do is simulate span traffic somewhat-similar to production shape traffic by generating synthetic traffic to multiple systems. So in my case, I'm adding multiple tracers (e.g., Instana and Jaeger) but also multiple services, so that I can visualize the same set of traces in multiple systems, but without actually starting lots of processes or lots of different nodes / services.

In my admittedly-weird use case, I noticed that the same spans that other systems would visualize as separate services were getting mapped all to one service. After talking to Instana support, I was under the (turns-out-wrong) impression, that the 'appname' (instana.appname) was the thing that resulted in different colors for different services, so I did just-enough of the globals-ectomy to get it sending different service names. This included making the sensor non-global and allowing you to set service_name in the Options object you pass to sensor.__init__ and not using global sensors.

pglombardo:
By all means, I appreciate the suggestions/PR.  We want the Python sensor to be as flexible as possible for all potential use cases.

Service naming is hard and definitely varies across vendors.  Hopefully that doc I linked explains our strategy well enough (and the 5 ways to name a service)...  It's still a draft though.

Yes the display & handling of Opentracing spans depends on the annotations on that span.  We put a section up recently to describe this.  It may help: https://docs.instana.io/ecosystem/open_tracing/#protip-best-use-of-opentracing-tags-for-better-processing

Keep sending the suggestions.  Many of the changes in this PR I was planning to do anyway along with the service naming overhaul for Python.

pglombardo:
Hi @nyanshak - we've removed much of the global state and service naming has been re-done with the new Application Perspectives feature.  I'm going to close this PR.  Thanks for contributing!

PR #82, opened 2018-06-25
No pid in the logger instance

pglombardo:
## Problem
When using custom logging is hard to override the handler and format of a logger that is named after a process id.

https://github.com/instana/python-sensor/blob/3810da69e05df351e54ca6685160acb9bd0a20f7/instana/log.py#L4

## Proposal
Use the LogRecord attributes to output a per process or per thread log in the format instead.
Optional, use the `dictConfig` to set up loggers.

### In
https://github.com/instana/python-sensor/blob/3810da69e05df351e54ca6685160acb9bd0a20f7/instana/log.py#L9
### Do
`[%(process)d - %(thread)d] ...`
and name the logger `instana`

## Reference
* https://docs.python.org/3/library/logging.html#logrecord-attributes
* https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig

### DIFF
--- a/instana/fsm.py
+++ b/instana/fsm.py
@@ -36,6 +36,8 @@ class Fsm(object):
     fsm = None
     timer = None
 
+    warnedPeriodic = False
+
     def __init__(self, agent):
         log.info("Stan is on the scene.  Starting Instana instrumentation version", instana.__version__)
         log.debug("initializing fsm")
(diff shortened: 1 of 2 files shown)


arnulfojr:
@pglombardo Thanks! Awesome, next time I'll do the PR! 😄

pglombardo:
This improvement has been released in 0.10.1: https://github.com/instana/python-sensor/releases/tag/0.10.1

PR #84, opened 2018-06-26
MySQLdb Instrumentation (mysql-python)

pglombardo:
New instrumentation for the MySQL-python package:

https://pypi.org/project/MySQL-python/


pglombardo:
This PR is replaced by https://github.com/instana/python-sensor/pull/87

PR #121, opened 2019-01-08
Clarify automatic Python instr needs static binary

mmanciop:


pglombardo:
This is actually not the correct place for this note.

The python instrumentation can be enabled in 3 ways:

1. magic remote instrumentation
2. via environment variable (no code changes)
3. manual import (code change)

This section is number 2.  I'll update the other section with a note.  FYI

PR #128, opened 2019-02-07
New logging instrumentation

steveww:
Additional instrumentation for standard logging library.

### DIFF
--- a/instana/instrumentation/logging.py
+++ b/instana/instrumentation/logging.py
@@ -12,32 +12,34 @@ def log_with_instana(wrapped, instance, argv, kwargs):
     # argv[0] = level
     # argv[1] = message
     # argv[2] = args for message
-    parent_span = tracer.active_span
+    try:
+        parent_span = tracer.active_span
 
-    # Only needed if we're tracing and serious log
-    if parent_span and argv[0] >= logging.WARN:
-        # get the formatted log message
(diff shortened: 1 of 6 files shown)


instana/instrumentation/logging.py
```diff
@@ -0,0 +1,43 @@
...
+import wrapt
+import logging
+import sys
+
+from ..log import logger
+from ..singletons import tracer
+
+@wrapt.patch_function_wrapper('logging', 'Logger._log')
+def log_with_instana(wrapped, instance, argv, kwargs):
+    # argv[0] = level
```
pglombardo: We should wrap this in a try/except/finally to protect against anything going wrong and not losing the log message for the end user.

instana/recorder.py
```diff
@@ -152,6 +152,17 @@ def build_registered_span(self, span):
                 tskey = list(data.custom.logs.keys())[0]
                 data.mysql.error = data.custom.logs[tskey]['message']
 
+        if span.operation_name == "log":
+            logger.debug('found a log')
```
pglombardo: Can remove this.

pglombardo:
Looks great @steveww!  The new Log span now broke some of the other tests (because of the extra span) but I'll fix those in a subsequent PR.

PR #131, opened 2019-02-15
fixed case

steveww:
The type field for the SDK data object had the wrong case Type not type. The trace processing on the backend does a case sensitive lookup of this field. It does not find Type so tags the span as entry, the default. This makes it impossible to have an intermediate span, this fixes that.

### DIFF
--- a/instana/json_span.py
+++ b/instana/json_span.py
@@ -124,7 +124,7 @@ class SDKData(object):
     name = None
     type = None
     arguments = None
-    Return = None
+    return = None
     custom = None
 
     def __init__(self, **kwds):
--- a/instana/json_span.py
+++ b/instana/json_span.py
@@ -122,7 +122,7 @@ class SoapData(object):
 
 class SDKData(object):
     name = None
-    Type = None
+    type = None
     arguments = None
     Return = None
     custom = None
--- a/instana/recorder.py
+++ b/instana/recorder.py
@@ -210,7 +210,7 @@ class InstanaRecorder(SpanRecorder):
         sdk_data = SDKData(name=span.operation_name,
                            custom=custom_data)
 
-        sdk_data.Type = self.get_span_kind(span)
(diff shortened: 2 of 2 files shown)


instana/json_span.py
```diff
@@ -122,7 +122,7 @@ def __init__(self, **kwds):
 
 class SDKData(object):
     name = None
-    Type = None
+    type = None
     arguments = None
     Return = None
```
mmanciop: Nice catch. Shouldn’t we do the same with `Return`? Also: I think we should fix it backend side too as this does not handle already compiled GoLang executables.

steveww: Good point I was so focused on the Type I missed Return. Just committed that too.
I'm just started to look at Golang to see if it has the same problems.
It would be quite a big change on the backend to make it case insensitive and could have performance implications due to the high volume of traces processed.

pglombardo:
At one point there was a reason why we couldn't lowercase this field: https://github.com/instana/python-sensor/commit/396a279b1974d3a4c48cbc4665a04cd1611d557c#diff-92b2d0cde86ecf3b3bb28fe5e95496d6

I'll take another look at this today.  Thanks for the PR!

steveww:
Peter,

Could be a Python 2.x thing? I only tested with Python 3.x


Regards,


Steve Waterworth
s@steveww.org

pglombardo:
No - the backend has evolved and we haven't followed.  Fixing this today.  I'll might have to create another branch/PR to make sure we properly report `k` like we just did for Go as well.

pglombardo:
Ah I remember now :-) lowercase `type` and `return` are Python keywords.  We'll have to lowercase in json encoding potentially.  I'll see what magic I can pull...

pglombardo:
This PR is replaced by https://github.com/instana/python-sensor/pull/132

Thanks for the PR @steveww!

PR #139, opened 2019-04-02
Asyncio: Carry context along with Tasks

pglombardo:
Carry context across ensure_future calls


pglombardo:
How to enable/disable this option: https://github.com/instana/python-sensor/blob/master/Configuration.md#package-configuration

PR #160, opened 2019-05-28
Console support

pglombardo:
This PR adds the ability to invoke an instana console with:

    python -m instana console

The console depends on ipython and is not installed by default with Instana.

To enable the console, you must install ipython.  Package output describes as much.


pglombardo:
Screen Shot 2019-05-28 at 3 09 15 PM
Screen Shot 2019-05-28 at 3 08 17 PM

PR #173, opened 2019-06-20
Fix for issue 172

steveww:
Leave the scope open for downstream error handles and logging instrumentation.

### DIFF
--- a/instana/instrumentation/flask/vanilla.py
+++ b/instana/instrumentation/flask/vanilla.py
@@ -74,16 +74,18 @@ def after_request_with_instana(response):
 @wrapt.patch_function_wrapper('flask', 'Flask.handle_user_exception')
 def handle_user_exception_with_instana(wrapped, instance, argv, kwargs):
     exc = argv[0]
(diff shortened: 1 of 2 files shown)


steveww:
New commit should ensure span is closed or left open if there is a downstream handler defined.

steveww:
I think that perhaps the wrapper around handle_user_exception should create a log span so that the error message gets picked up under Log Messages in the UI. WDYT?

steveww:
Any update on this?

pglombardo:
Hi Steve - this probably won't progress until after August 1.  

Your use of `._find_error_handler` is interesting.  Thus far I prefer the approach to fix in #172 but I will test both again when I revisit this.  Need to validate the edge cases etc..

steveww:
Peter,

Any chance you have time to dig into this and review it?

pglombardo:
Thanks for this @steveww - I put together #172 which has a few more fixes improvements too.  Going to close this in favor of #172.

PR #176, opened 2019-06-27
Psycopg2: Assure index exists before referencing

pglombardo:
Also: Make sure not to send empty trace payloads


pglombardo:
PR was pushed with the wrong title - hub caching issue

PR #207, opened 2019-12-18
uWSGI postfork hooks

pglombardo:
* Add uWSGI postfork hook support


pglombardo:
Tested on Python versions 2.7.17 -> 3.7.5 with and without `--lazy-apps` and again with and without `--lazy`.  All cases announce works properly.

This PR removes the requirement for the `--lazy-apps` option as in the docs that will be updated when this is released.

PR #231, opened 2020-05-05
Add support for X-INSTANA-SYNTHETIC

mmanciop:


pglombardo:
Thanks @mmanciop.  There are two issues:

1. We need to modify [BaseSpan](https://github.com/instana/python-sensor/blob/0b1586e52d2dac0e7964de67f49759e945b78a24/instana/span.py#L125) instead of adding `sy` as a tag.
2. The header detection should be done in the HTTP propagator

No worries though - I'll pick this up and make the changes once I get some cycles.  Thanks!

PR #269, opened 2020-09-25
Instrument Google Cloud Storage client

andrewslotin:
This PR adds instrumentation for [`google-cloud-storage`](https://googleapis.dev/python/storage/latest/index.html) client by wrapping the `google.cloud.storage._http.Connection.api_request()` and extracting span data from the request made to Google Cloud Storage JSON API.

Since `storage.Client` does not use this API to upload and download blobs, additional wrappers are needed to instrument `storage.blob.Blob.{upload,download}_*` calls.

Batches are implemented as a wrapper for `storage._http.Connection` and thus also needed a special handling.


pglombardo (approved):
Nice work @andrewslotin

PR #279, opened 2020-10-16
AWS Lambda Safeties and Improvements

pglombardo:


instana/instrumentation/aws/lambda_inst.py
```diff
@@ -31,11 +31,6 @@ def lambda_handler_with_instana(wrapped, instance, args, kwargs):
                     result['headers']['Server-Timing'] = server_timing_value
                 elif 'multiValueHeaders' in result:
                     result['multiValueHeaders']['Server-Timing'] = [server_timing_value]
-                else:
-                    # If both 'headers' and 'multiValueHeaders' aren't in result,
-                    # then default to setting single value 'headers'
-                    result['headers'] = dict()
-                    result['headers']['Server-Timing'] = server_timing_value
```
basti1302, line 38: @pglombardo I disagree with removing this entirely just for the edge case of Authorizers. It will break website monitoring back end correlation for a lot of customers. See https://instana.slack.com/archives/GBS2G15GD/p1602840352151100 for details.

basti1302 (changes requested):

instana/instrumentation/aws/lambda_inst.py
pglombardo: As mentioned in Slack, Authorizers are called by the API gateway to validate incoming requests and have a strict response format.  This will not break website correlation as it doesn't apply in this case.  (API Gateway responses will still return context) We can revisit this next week if needed.

basti1302: I don't follow. If I read this change correctly, the Python Lambda wrapper will no longer add the server timing header when the wrapped Lambda function returns a response object which has neither the `headers` nor the `multiValueHeaders` attribute. 

Where in this code do you make sure that this only applies to authorizers?

PR #288, opened 2020-11-18
Add support for API Gateway v2 Lambda payload

andrewslotin:
AWS has [introduced](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html) a new trigger event type, which is now a default for newly created triggers. This PR adds support for such payload format, extracting the trace context and necessary tags for the entry span.

### DIFF
--- a/instana/instrumentation/aws/triggers.py
+++ b/instana/instrumentation/aws/triggers.py
@@ -18,7 +18,7 @@ def get_context(tracer, event):
         is_application_load_balancer_trigger(event)
 
     if is_proxy_event:
-        return tracer.extract('http_headers', event['headers'])
+        return tracer.extract('http_headers', event.get('headers', {}))
 
     return tracer.extract('http_headers', event)
 
--- a/instana/instrumentation/aws/triggers.py
+++ b/instana/instrumentation/aws/triggers.py
@@ -13,7 +13,11 @@ STR_LAMBDA_TRIGGER = 'lambda.trigger'
 
 def get_context(tracer, event):
     # TODO: Search for more types of trigger context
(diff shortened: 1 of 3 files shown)


instana/instrumentation/aws/triggers.py
```diff
@@ -13,7 +13,11 @@
 
 def get_context(tracer, event):
     # TODO: Search for more types of trigger context
-    if is_api_gateway_proxy_trigger(event) or is_application_load_balancer_trigger(event):
+    is_proxy_event = is_api_gateway_proxy_trigger(event) or \
+        is_api_gateway_v2_proxy_trigger(event) or \
+        is_application_load_balancer_trigger(event)
+
+    if is_proxy_event:
         return tracer.extract('http_headers', event['headers'])
```
pglombardo, line 21: The failsafe check here would be to also add in `if "headers" in event` or alternatively wrap this entire function in an try..except (or both).  What do you think?

instana/instrumentation/aws/triggers.py
```diff
@@ -26,6 +30,20 @@ def is_api_gateway_proxy_trigger(event):
     return True
 
 
+def is_api_gateway_v2_proxy_trigger(event):
```
pglombardo, line 33: Nice

pglombardo (commented):
This is a great improvement - thanks

instana/instrumentation/aws/triggers.py
andrewslotin: Fair enough, although the fact that this is a proxy event guarantees that we have `headers`, who knows when AWS decides to change their payload again. An `event.get('headers', dict())` maybe?

PR #294, opened 2021-02-22
Do not mark Django HTTP 404 response spans as errors

andrewslotin:
This PR updates Django instrumentation to not to count `django.http.response.Http404` exceptions as errors. This is a perfectly normal case for an app to signal that the requested document is not found, hence it should not be reported as an error.

### DIFF
--- a/instana/instrumentation/django/middleware.py
+++ b/instana/instrumentation/django/middleware.py
@@ -59,7 +59,6 @@ class InstanaMiddleware(MiddlewareMixin):
                 request.iscope.span.set_tag(ext.HTTP_STATUS_CODE, response.status_code)
                 tracer.inject(request.iscope.span.context, ot.Format.HTTP_HEADERS, response)
                 response['Server-Timing'] = "intid;desc=%s" % request.iscope.span.context.trace_id
-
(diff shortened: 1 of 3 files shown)


instana/instrumentation/django/middleware.py
```diff
@@ -69,6 +68,11 @@ def process_response(self, request, response):
             return response
 
     def process_exception(self, request, exception):
+        from django.http.response import Http404
```
manojpandey, line 71: nitpick: can we drop this to the top of the file where all the imports are?

manojpandey (approved):
LGTM, except a small minor comment.

instana/instrumentation/django/middleware.py
andrewslotin: The idea is that we import this class in a place where we're sure that django is present. Alternatively I could do this at the beginning of `load_middleware_wrapper()` where we are also importing settings. WDYT?

manojpandey: No, then it's okay - LGTM 👍

manojpandey (approved):

PR #295, opened 2021-02-26
Fix CI build

andrewslotin:
This PR addresses multiple issues that cause tests to fail on CI:
* `moto` has released version 2.0, that is incompatible with Python 2.7. The library version is now locked to the latest 1.3 release
* `google-cloud-storage` has added mtls support in 1.36.0. I've updated the HTTP request session mock to address this change
* `rsa` has dropped support for Python 2.7 as of version 4.6. The test suite now uses a separate requirements list for this Python version, where `rsa` is locked to v4.5
* Python 3 builds now use the latest version of `pip` to avoid building `cryptography` (requires Rust compiler)


tests/clients/test_google-cloud-storage.py
```diff
@@ -10,7 +10,7 @@
 from instana.singletons import tracer
 from ..test_utils import _TraceContextMixin
 
-from mock import patch, Mock
+from mock import patch, Mock, create_autospec
```
manojpandey, line 13: Are you using this `create_autospec` anywhere?

manojpandey (approved):
Sounds good to me, thanks @andrewslotin !

tests/clients/test_google-cloud-storage.py
andrewslotin: Good catch, this is a leftover from my changes. Will update

Issue #303, opened 2021-03-19
Sqlalchemy will drop support for Python 2 after v.1.4


manojpandey:
https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#python-3-6-is-the-minimum-python-3-version-python-2-7-still-supported

manojpandey:
We lock the version in this PR: https://github.com/instana/python-sensor/pull/304/commits/be5478108da8f6a4a2086f4b8760d9b8f09ee711

manojpandey:
Closed with https://github.com/instana/python-sensor/pull/304

PR #304, opened 2021-03-19
Fix: Make CI happy, sqlalchemy dbapi_error update

manojpandey:
Due to Sqlalchemy updates in v.1.4 the event `dbapi_error` does not exists anymore.
```
ImportError while loading conftest '/home/circleci/repo/tests/conftest.py'.
tests/conftest.py:45: in <module>
    import instana
instana/__init__.py:198: in <module>
    boot_agent()
instana/__init__.py:166: in boot_agent
    from .instrumentation import sqlalchemy
instana/instrumentation/sqlalchemy.py:50: in <module>
    @event.listens_for(Engine, 'dbapi_error', named=True)
venv/lib/python2.7/site-packages/sqlalchemy/event/api.py:142: in decorate
    listen(target, identifier, fn, *args, **kw)
venv/lib/python2.7/site-packages/sqlalchemy/event/api.py:99: in listen
    _event_key(target, identifier, fn).listen(*args, **kw)
venv/lib/python2.7/site-packages/sqlalchemy/event/api.py:30: in _event_key
    "No such event '%s' for target '%s'" % (identifier, target)
E   InvalidRequestError: No such event 'dbapi_error' for target '<class 'sqlalchemy.engine.base.Engine'>'
```

until v.1.3 there were both two events recognised: `dbapi_error` and the `handle_error` event
the signatures were
- `handle_error(self, exception_context):`
- `dbapi_error(self, conn, cursor, statement, parameters, context, exception)`

starting v.1.4, `dbapi_error` is gone now and hence throws:
```error
InvalidRequestError: No such event 'dbapi_error' for target '<class 'sqlalchemy.engine.base.Engine'>'
```

This PR adds a way to handle both the events.

To keep in mind for future: https://github.com/instana/python-sensor/issues/303 
This PR also locks the version to 1.4 in case of Python 2.x

### DIFF
--- a/instana/instrumentation/sqlalchemy.py
+++ b/instana/instrumentation/sqlalchemy.py
@@ -58,15 +58,15 @@ try:
 
     @event.listens_for(Engine, error_event, named=True)
     def receive_handle_db_error(**kw):
-        context = kw['context']
+        context = kw['exception_context'].execution_context
 
         if context is not None and hasattr(context, '_stan_scope'):
             scope = context._stan_scope
             if scope is not None:
                 scope.span.mark_as_errored()
 
-                if 'exception' in kw:
-                    e = kw['exception']
+                if context.exception:
+                    e = context.exception
                     scope.span.set_tag('sqlalchemy.err', str(e))
                 else:
                     scope.span.set_tag('sqlalchemy.err', "No %s specified." % error_event)
--- a/instana/instrumentation/sqlalchemy.py
+++ b/instana/instrumentation/sqlalchemy.py
@@ -15,6 +15,7 @@ try:
 
     url_regexp = re.compile(r"\/\/(\S+@)")
 
+
     @event.listens_for(Engine, 'before_cursor_execute', named=True)
     def receive_before_cursor_execute(**kw):
         try:
@@ -38,6 +39,7 @@ try:
         finally:
             return
 
+
     @event.listens_for(Engine, 'after_cursor_execute', named=True)
     def receive_after_cursor_execute(**kw):
         context = kw['context']
@@ -47,8 +49,15 @@ try:
             if scope is not None:
                 scope.close()
 
-    @event.listens_for(Engine, 'dbapi_error', named=True)
-    def receive_dbapi_error(**kw):
+
+    error_event = "handle_error"
+    # Handle dbapi_error event; deprecated since version 0.9
+    if sqlalchemy.__version__[0] == "0":
+        error_event = "dbapi_error"
+
+
+    @event.listens_for(Engine, error_event, named=True)
+    def receive_handle_db_error(**kw):
         context = kw['context']
 
         if context is not None and hasattr(context, '_stan_scope'):
@@ -60,7 +69,7 @@ try:
                     e = kw['exception']
                     scope.span.set_tag('sqlalchemy.err', str(e))
                 else:
-                    scope.span.set_tag('sqlalchemy.err', "No dbapi error specified.")
+                    scope.span.set_tag('sqlalchemy.err', "No %s specified." % error_event)
                 scope.close()
 
 
--- a/instana/instrumentation/sqlalchemy.py
+++ b/instana/instrumentation/sqlalchemy.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import
 
 import re
+from operator import attrgetter
 
 from ..log import logger
 from ..singletons import tracer
@@ -14,7 +15,6 @@ try:
     from sqlalchemy.engine import Engine
 
     url_regexp = re.compile(r"\/\/(\S+@)")
-    logger.debug("Instrumenting sqlalchemy")
 
 
     @event.listens_for(Engine, 'before_cursor_execute', named=True)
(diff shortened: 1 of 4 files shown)


instana/instrumentation/sqlalchemy.py
```diff
@@ -47,8 +49,18 @@ def receive_after_cursor_execute(**kw):
...
-    @event.listens_for(Engine, 'dbapi_error', named=True)
-    def receive_dbapi_error(**kw):
+
+    # Deprecated since version 0.9: The ConnectionEvents.dbapi_error() event is
+    # deprecated and will be removed in a future release. Please refer to the
+    # ConnectionEvents.handle_error() event.
+    if sqlalchemy.__version__[0] != "0":
+        error_event = "handle_error"
+    else:
+        error_event = "dbapi_error"
```
andrewslotin, line 59: A minor note: I believe it's safer (and slightly more readable) to inititalize the value first, and update it if needed:
```python
error_event = "handle_error"

# Handle 0.x explicitly
if sqlalchemy.__version__[0] == "0":
  error_event = "dbapi_error"
```

instana/instrumentation/sqlalchemy.py
```diff
@@ -60,7 +72,7 @@ def receive_dbapi_error(**kw):
                     e = kw['exception']
                     scope.span.set_tag('sqlalchemy.err', str(e))
                 else:
-                    scope.span.set_tag('sqlalchemy.err', "No dbapi error specified.")
+                    scope.span.set_tag('sqlalchemy.err', "No handle error specified.")
```
andrewslotin, line 75: Shall we use `error_event` here instead of hardcoded string?

andrewslotin (changes requested):
Since we're addressing SQLAlchemy upgrade, could you lock the max version for 2.7 too?

instana/instrumentation/sqlalchemy.py
manojpandey: 👍 done!

manojpandey:
Requested changes done, let's wait for the CI

instana/instrumentation/sqlalchemy.py
```diff
@@ -47,20 +49,27 @@ def receive_after_cursor_execute(**kw):
...
-            scope = context._stan_scope
+    error_event = "handle_error"
+    # Handle dbapi_error event; deprecated since version 0.9
+    if sqlalchemy.__version__[0] == "0":
+        error_event = "dbapi_error"
+
+
+    @event.listens_for(Engine, error_event, named=True)
+    def receive_handle_db_error(**kw):
+        context = kw['exception_context']
```
andrewslotin, line 61: But it will be `'context'` for older versions, right? Shall we fall back to this value?

andrewslotin (changes requested):

instana/instrumentation/sqlalchemy.py
```diff
@@ -47,21 +49,43 @@ def receive_after_cursor_execute(**kw):
...
+
+                if scope is not None:
+                    scope.span.mark_as_errored()
+                    if context.exception:
+                        e = context.exception
+                        scope.span.set_tag('sqlalchemy.err', str(e))
+                    else:
+                        scope.span.set_tag('sqlalchemy.err', "No %s specified." % error_event)
+                    scope.close()
+        else:
```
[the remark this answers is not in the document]
manojpandey: update in the latest revision :D

instana/instrumentation/sqlalchemy.py
```diff
@@ -47,23 +50,43 @@ def receive_after_cursor_execute(**kw):
...
+            context = kw.get('exception_context')
+            if hasattr(context.execution_context, '_stan_scope') and \
+                    hasattr(context, 'sqlalchemy_exception'):
+                scope = context.execution_context._stan_scope
+                context_exception = context.sqlalchemy_exception
+                if scope:
+                    _set_error_tags(scope, context_exception)
 
-    logger.debug("Instrumenting sqlalchemy")
```
andrewslotin, line 67: The idea for this log line to be the last is that we write it once instrumentation has succeeded. Should there be an exception during initialization, there will be no such line in the log, which gives us a hint that there was a problem.

instana/instrumentation/sqlalchemy.py
```diff
@@ -47,23 +50,43 @@ def receive_after_cursor_execute(**kw):
...
+                scope = context._stan_scope
+                context_exception = context.exception
+                if scope:
+                    _set_error_tags(scope, context_exception)
+
+        else:
+            context = kw.get('exception_context')
+            if hasattr(context.execution_context, '_stan_scope') and \
+                    hasattr(context, 'sqlalchemy_exception'):
+                scope = context.execution_context._stan_scope
```
andrewslotin, line 86: Would it make more sense to pass the context instead of a scope to the `_set_error_tags()` and have the `hasattr(context, '_stan_scope')` there to reduce duplication?

instana/instrumentation/sqlalchemy.py
```diff
@@ -47,23 +50,43 @@ def receive_after_cursor_execute(**kw):
...
+                context_exception = context.sqlalchemy_exception
+                if scope:
```
andrewslotin, line 88: I believe this could be moved outside of condition, since it's common for both `if` and `else`.

instana/instrumentation/sqlalchemy.py
```diff
@@ -47,23 +50,43 @@ def receive_after_cursor_execute(**kw):
...
 
-                if 'exception' in kw:
-                    e = kw['exception']
-                    scope.span.set_tag('sqlalchemy.err', str(e))
-                else:
-                    scope.span.set_tag('sqlalchemy.err', "No dbapi error specified.")
-                scope.close()
 
+    def _set_error_tags(scope, exception):
+        scope.span.mark_as_errored()
```
andrewslotin, line 61: I know this goes a bit outside of current PR, since this was how it's been implemented originally, but usually we call `InstanaSpan.log_exception()` to add exceptions to the span. You might want to add a special handling for SQLAlchemy operation there though.

andrewslotin (changes requested):

manojpandey:
Hey @andrewslotin - Updated the PR with the requested changes. It was a bit tricky to write as less code as possible because of the different exception attributes and the scope attributes. Fortunately, this [operator.attrgetter](https://docs.python.org/3/library/operator.html#operator.attrgetter) came very handy (both in Python 2/3).

andrewslotin (approved):
👍

PR #309, opened 2021-04-13
Fix: RabbitMQ check proper span type

### LINKED ISSUE #301: Check implication of __init__ for rabbitmq span type
Specifically this piece of code: https://github.com/instana/python-sensor/blob/f800501fe1259322056bf1f2bbb0516de688e9a7/instana/span.py#L257-L258

```
    def __init__(self, span, source, service_name, **kwargs):
        # pylint: disable=invalid-name
        super(RegisteredSpan, self).__init__(span, source, service_name, **kwargs)
        self.n = span.operation_name
        self.k = 1
        if span.operation_name in self.ENTRY_SPANS:
            # entry
            self._populate_entry_span_data(span)
            self.data["service"] = service_name
        elif span.operation_name in self.EXIT_SPANS:
            self.k = 2 # exit
            self._populate_exit_span_data(span)
        elif span.operation_name in self.LOCAL_SPANS:
            self.k = 3 # intermediate span
            self._populate_local_span_data(span)

        if "rabbitmq" in self.data and self.data["rabbitmq"]["sort"] == "consume":
            self.k = 1  # entry
```

Looks like, it has no effect.

pdimitra:
Closes #301

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,6 +76,30 @@ jobs:
             . venv/bin/activate
             pytest -v
 
+  python37:
+    docker:
+      - image: circleci/python:3.7.9
+      - image: circleci/postgres:9.6.5-alpine-ram
+      - image: circleci/mariadb:10-ram
+      - image: circleci/redis:5.0.4
+      - image: rabbitmq:3.5.4
+      - image: circleci/mongo:4.2.3-ram
+      - image: singularities/pubsub-emulator
+        environment:
+          PUBSUB_PROJECT_ID: "project-test"
+          PUBSUB_LISTEN_ADDRESS: "0.0.0.0:8432"
+    working_directory: ~/repo
+    steps:
+      - checkout
+      - pip-install-deps
+      - run:
+          name: run tests
+          environment:
+            INSTANA_TEST: "true"
+          command: |
+            . venv/bin/activate
+            pytest -v
+
   python38:
     docker:
       - image: circleci/python:3.8.6
@@ -227,6 +251,7 @@ workflows:
   build:
     jobs:
       - python27
+      - python37
       - python38
       - python39
       - py27cassandra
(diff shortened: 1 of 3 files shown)


instana/span.py
```diff
@@ -247,24 +248,23 @@ def __init__(self, span, source, service_name, **kwargs):
...
         elif span.operation_name in self.EXIT_SPANS:
-            self.k = 2 # exit
+            self.k = 2  # exit
             self._populate_exit_span_data(span)
         elif span.operation_name in self.LOCAL_SPANS:
-            self.k = 3 # intermediate span
+            self.k = 3  # intermediate span
             self._populate_local_span_data(span)
 
-        if "rabbitmq" in self.data and self.data["rabbitmq"]["sort"] == "consume":
```
manojpandey, line 261: @pdimitra:

I checked `span.py` and it seems like the `_populate_entry_span_data` & `_populate_exit_span_data` for `rabbitmq` is the same. In this case, we can let it stay in the `ENTRY_SPANS` tuple and just change the method here to `publish` and set `self.k = 2`

instana/span.py
```diff
@@ -237,7 +238,7 @@ class RegisteredSpan(BaseSpan):
                   "mongo", "mysql", "postgres", "rabbitmq", "redis", "rpc-client", "sqlalchemy",
                   "soap", "tornado-client", "urllib3", "pymongo", "gcs", "gcps-producer")
 
-    ENTRY_SPANS = ("aiohttp-server", "aws.lambda.entry", "celery-worker", "django", "wsgi", "rabbitmq",
```
manojpandey, line 240: We can keep this. See my other comment below

.circleci/config.yml
```diff
@@ -76,6 +76,30 @@ jobs:
...
+      - checkout
+      - pip-install-deps
+      - run:
+          name: run tests
+          environment:
+            INSTANA_TEST: "true"
+          command: |
+            . venv/bin/activate
+            pytest -v
+
```
manojpandey, line 102: Let's see if we can run `asqynp` on Python 3.8 - if not, then we can have this additional test job.

manojpandey (changes requested):
@pdimitra thanks for working on this - I added some comments in the review :D 
Good work!

manojpandey (approved):
LGTM!

.circleci/config.yml
manojpandey: Looks like support only until Python 3.7: https://github.com/benjamin-hodgson/asynqp/pull/104 
We can keep it @pdimitra

PR #311, opened 2021-05-05
Add coverage report

pdimitra:


.circleci/config.yml
```diff
@@ -122,7 +122,11 @@ jobs:
             INSTANA_TEST: "true"
           command: |
             . venv/bin/activate
-            pytest -v
+            coverage run --source=instana -m pytest -v
+            coverage report -m
+            coverage html
+      - store_artifacts:
+          path: htmlcov
```
andrewslotin, line 129: What would be the use case for this artifact?

pdimitra: This makes the html coverage report accessible in the circleci artifacts of the specific job https://circleci.com/docs/2.0/code-coverage/. I only used the coverage package in python to do it, not codecov or any other tool

andrewslotin: Right, but it seems like we have the same information in the CI build log that is also stored with the build. Do you think we need these HTML files as well? What would be the use pattern for it?

pdimitra: The is no additional information. It is just a different visualisation, as in the html report you can see the actual code which is not tested, without needing to open your IDE and look for the lines at this point. This is the only benefit I can see. Personally I was alway using the info from the build, Manoj if I remember right liked the html visualisation. It is a matter of taste, I am fine with any of them. I can take the html report out if you think it has no usability.

andrewslotin: Alright, I see your point now 👍 If there is a use in these reports, let's keep them!

andrewslotin (approved):

PR #313, opened 2021-05-18
Instrument sanic

pdimitra:


manojpandey (changes requested):
@pdimitra To make sure we only run the tests for Sanic, please add it here in conftest.py

https://github.com/instana/python-sensor/blob/57a79dff61b501314d91aa8299de3138b3fbebc2/tests/conftest.py#L37 
```python
collect_ignore_glob.append("*test_sanic*")
```

instana/instrumentation/sanic_inst.py
```diff
@@ -0,0 +1,138 @@
...
+        try:
+            request = args[0]
+            try:  # scheme attribute is calculated in the sanic handle_request method for v19, not yet present
+                if "http" not in request.scheme:
+                    return await wrapped(*args, **kwargs)
+            except AttributeError:
+                pass
+            headers = request.headers.copy()
+            ctx = async_tracer.extract(opentracing.Format.HTTP_HEADERS, headers)
+            with async_tracer.start_active_span("asgi", child_of=ctx) as tracing_scope:
```
manojpandey, line 105: Can you change `tracing_scope` to `scope` here as well as in `asgi.py` - seems like we're using `scope` in most places for consistency

instana/instrumentation/sanic_inst.py
```diff
@@ -0,0 +1,138 @@
...
+                if isinstance(query, (str, bytes)) and len(query):
+                    if isinstance(query, bytes):
+                        query = query.decode('utf-8')
+                    scrubbed_params = strip_secrets_from_query(query, agent.options.secrets_matcher,
+                                                               agent.options.secrets_list)
+                    tracing_scope.span.set_tag("http.params", scrubbed_params)
+
+                if agent.options.extra_http_headers is not None:
+                    extract_custom_headers(tracing_scope, headers)
+                await wrapped(*args, **kwargs)
```
manojpandey, line 122: Shall we also check the response and set `scope.span.mark_as_errored()`?

instana/instrumentation/sanic_inst.py
```diff
@@ -0,0 +1,138 @@
...
+                tracing_scope.span.set_tag("http.url", request.url)
+                if hasattr(request, "uri_template"):
+                    tracing_scope.span.set_tag("http.path_tpl", request.uri_template)
+                if hasattr(request, "ctx"):  # ctx attribute added in the latest v19 versions
+                    request.ctx.iscope = tracing_scope
```
manojpandey, line 127: Can we put them together around L106-109?

instana/instrumentation/sanic_inst.py
```diff
@@ -0,0 +1,138 @@
...
+        except Exception as e:
+            logger.debug("Sanic framework @ process_request", exc_info=True)
```
manojpandey, line 129: `process_request` or `handle_request`? (`Sanic.handle_request`)

manojpandey (changes requested):
Thanks, @pdimitra - this is great. I've added some comments!

tests/requirements.txt
```diff
@@ -23,6 +23,7 @@ pyramid>=2.0
 pytest>=6.2.4
 pytest-celery
 redis>=3.5.3
+sanic>=19.0.0
```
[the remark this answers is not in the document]
pdimitra: The actual instrumentation is tested from the 19 up, because the customer uses 19.03.01. A couple of if blocks could be cleaned up if we were going for 20.12, but I suppose we need customer feedback

instana/instrumentation/sanic_inst.py
pdimitra: `"uri_template"` is only getting evaluated after the `await wrapped(*args, **kwargs)` line 122.
Line 123 I can move up. Lines 126-127 could go up, but it can be tricky for early v19 versions, as adding more things afterwards to the scope will require and update or inplace update of the request attribute if ctx exists. Additionally attaching the scope metrics to the request before executing the `await wrapped(*args, **kwargs)` exposes it to the customer, who may also use the ctx attribute for his own purpose and we should avoid him potentially cleaning the existing data. Do we have a reason for doing this? I suppose, if we want to give our already calculated tracing information to the customer for him to interact programatically, we could, but we should provide the customer with documentation of what is there and how to handle it.

instana/instrumentation/sanic_inst.py
```diff
@@ -0,0 +1,138 @@
...
+                span.log_exception(exc)
+        else:
+            wrapped(*args, **kwargs)
+
+
+    def response_details(span, response):
+        try:
+            status_code = response.status
+            if status_code is not None:
+                if 500 <= int(status_code) <= 511:
```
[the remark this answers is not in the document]
pdimitra: This block of code is common in all existing web frameworks instrumentations. Instana does not handling 4xx as "errors" not server errors

instana/instrumentation/sanic_inst.py
pdimitra: At this point there is no access to the response, however the response is checked with one of the other patched methods, either `send` or `write_response` or `stream_response` depending on the sanic version. Additionally if there was 5xx status code the `exception_with_instana` method will add the error message if present to our span data. We can have a short look together in the flow as it might not be that clear.

tests/frameworks/test_sanic.py
```diff
@@ -0,0 +1,371 @@
...
+    assert ('http.error' not in asgi_span.data['sdk']['custom']['tags'])
+    assert ('http.params' not in asgi_span.data['sdk']['custom']['tags'])
+
+    assert (asgi_span.sy)
+    assert (urllib3_span.sy is None)
+    assert (test_span.sy is None)
+
+
+def test_custom_header_capture(server):
+    from instana.singletons import agent
```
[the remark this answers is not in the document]
pdimitra: No, I will try all your suggestions on the tests. 90% of the code in the tests is a copy paste of the fast_api ones, so maybe a bit outdated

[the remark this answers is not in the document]
manojpandey: No worries :D

manojpandey (approved):
thanks @pdimitra - LGTM!

PR #320, opened 2021-05-31
not setting the path_tpl at all when it evaluates to None

pdimitra:
This is the only change asked by Cedric, can go with any other future release


manojpandey (approved):
LGTM!

PR #321, opened 2021-06-10
added async_tracer support for pymongo instrumentation

pdimitra:
…pymongo instrumentation


manojpandey (approved):
LGTM, thanks @pdimitra

instana/instrumentation/pymongo.py
```diff
@@ -17,12 +17,15 @@ def __init__(self):
...
             parent_span = tracer.active_span
-
+            active_tracer = tracer
             # return early if we're not tracing
             if parent_span is None:
-                return
+                parent_span = async_tracer.active_span
+                if parent_span is None:
+                    return
+                active_tracer = async_tracer
```
andrewslotin, line 26: Nice trick! Would it make sense to extract the creation of an exit span into a helper method, that would check whether there is a parent span in any of two tracers?

pdimitra: Yes, I will give it a try already, since that part of code will need to be applied in all other clients/dbs too.

PR #324, opened 2021-06-16
fix the object parameter of hasattr

### LINKED ISSUE #323: Not passing command line arguments to the script
I am trying to setup instana according to: https://www.instana.com/docs/ecosystem/python/ where I:
1. pip install instana
2. export AUTOWRAPT_BOOTSTRAP=instana

Everything seems to be working except that I am no longer able to pass command line arguments to python scripts.

Given a script file test.py:
```
import sys
print(sys.argv)
```
when I run from the command line:
```
python test.py a b c
```
the output is empty list instead of passed arguments.

If I don't set: AUTOWRAPT_BOOTSTRAP=instana instana is not starting and everything is working fine.

python version is 3.8 run as a virtualenv

manojpandey:
`hasattr` takes an [object](https://docs.python.org/3/library/functions.html#hasattr) and string as a parameter

Fixes #323

### DIFF
--- a/instana/__init__.py
+++ b/instana/__init__.py
@@ -54,7 +54,7 @@ def load(_):
     environment variable.
     """
     # Work around https://bugs.python.org/issue32573
-    if not hasattr("sys", "argv"):
+    if not hasattr(sys, "argv"):
         sys.argv = ['']
     return None
 


andrewslotin (approved):
👍

PR #325, opened 2021-06-16
Fixed invalid use of hasattr

cotterpl:
Fix for https://github.com/instana/python-sensor/issues/323


manojpandey:
Hi @cotterpl - we just fixed this and released a new version here: https://pypi.org/project/instana/1.33.4/

PR #329, opened 2021-06-22
Enable client async and mixed tracing

pdimitra:
There are 3 tracers in python, the sync one, the async and the tornado, right now all the db clients or urllib instrumentations are checking only if there is a parent active span coming from the sync tracer, this is not the case when the parent is using the async_tracer or tornado tracer and in those cases we skip the tracing.
The implementation here intends to always get the current active tracer.


tests/clients/boto3/test_boto3_s3.py
```diff
@@ -19,7 +19,7 @@
 def setup_method():
     """ Clear all spans before a test run """
     tracer.recorder.clear_spans()
-    os.remove(download_target_filename)
+#    os.remove(download_target_filename)
```
andrewslotin, line 22: Is this line still needed?

tests/frameworks/test_sanic.py
```diff
@@ -92,7 +92,7 @@ def test_404(self):
         self.assertEqual(result.status_code, 404)
 
         spans = tracer.recorder.queued_spans()
-        self.assertEqual(len(spans), 3)
+        self.assertEqual(len(spans), 4)
```
andrewslotin, line 95: Out of curiosity, what caused this change of behavior?

instana/util/traceutils.py
```diff
@@ -12,18 +12,20 @@ def extract_custom_headers(tracing_scope, headers):
...
+            return tracer, tracer.active_span
         elif async_tracer.active_span:
-            return async_tracer
+            return async_tracer, async_tracer.active_span
+        elif tornado_tracer.active_span:
+            return tornado_tracer, tornado_tracer.active_span
         else:
-            return None
-    except Exception as e:
+            return None, None
```
andrewslotin, line 28: The name of this method suggests that it returns the active tracer, but now it also returns the active span. Considering that all tracers implement the same interface to obtain the active span, what was the reason behind changing the method signature and making it return the span as well?

Looking at how this method is being used, I see the following pattern:

```python
tracer, parent_span = get_active_tracer()

if parent_span is None:
    return
```

The `parent_span is None` is somehow contains a hidden knowledge: if there is no active span, it means we do not trace. How about we keep this method returning tracer only and change the parent span check to `if active_tracer is None`? I believe this way it'd more straightforward to understand why do we skip the tracing. This would also allow us to avoid throwing an exception if we accidentally call a method on `tracer` before checking the `parent_span`

instana/instrumentation/celery/hooks.py
```diff
@@ -55,8 +56,11 @@ def task_prerun(*args, **kwargs):
             headers = task.request.get('headers', {})
             if headers is not None:
                 ctx = tracer.extract(opentracing.Format.HTTP_HEADERS, headers)
+                active_tracer = tracer
+            else:
+                active_tracer, parent_span = get_active_tracer()
 
-            scope = tracer.start_active_span("celery-worker", child_of=ctx)
+            scope = active_tracer.start_active_span("celery-worker", child_of=ctx)
```
andrewslotin, line 63: I guess this will throw an exception if there is no active tracer and `headers` are `None`, right?

instana/instrumentation/celery/hooks.py
```diff
@@ -55,8 +56,11 @@ def task_prerun(*args, **kwargs):
             headers = task.request.get('headers', {})
             if headers is not None:
                 ctx = tracer.extract(opentracing.Format.HTTP_HEADERS, headers)
```
andrewslotin, line 58: We use the regular tracer to extract headers, even though we can switch to an async one to create a span later. Would it make sense to use the same instance all the time? I believe this would also address my comment below: if we get the active tracer early and fall back to the regular one if there is none, we'd know for sure that it's not `None`

andrewslotin (changes requested):

instana/autoprofile/runtime.py
```diff
@@ -3,6 +3,7 @@
 
 import sys
 import signal
+import os
```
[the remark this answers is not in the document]
pdimitra: I added the import because I am going also a bit through the files and checking the complaints in pycharm. It is complaining of unresolved import if I take it out because of line 42. `os.kill(os.getpid(), signum)`

tests/frameworks/test_sanic.py
pdimitra: there is an extra log span coming up after the changes, in the error test cases there are some logs and after the async_tracer is also available for the logging instrumentation, the extra span appears here

tests/clients/boto3/test_boto3_s3.py
pdimitra: I cleaned up all my changes in the boto3 tests, they were related to a piece of code I was trying and had reverted already, no need of changes there, only formatting changes on boto3 tests

andrewslotin (approved):
Excellent!

PR #335, opened 2021-09-08
1.40 release of grpcio is no longer supporting python27

pdimitra:
Master python27 pipeline is failing because it tries to build grpcio 1.40 in python2.7 which is no longer supported


andriikushch (approved):
Looks good to me.

PR #338, opened 2021-09-30
Sanic exception bug fix

pdimitra:
Fixing bug on evaluating args[0] even when message was present as a keyword argument and was causing an IndexError
Additional added a try/catch block to log any possible exception could be triggered by a bug in the instana code, so that this will not influence the execution flow of the customer's code.

### DIFF
--- a/instana/instrumentation/sanic_inst.py
+++ b/instana/instrumentation/sanic_inst.py
@@ -17,17 +17,21 @@ try:
 
     @wrapt.patch_function_wrapper('sanic.exceptions', 'SanicException.__init__')
     def exception_with_instana(wrapped, instance, args, kwargs):
-        message = kwargs.get("message", args[0])
-        status_code = kwargs.get("status_code")
-        span = async_tracer.active_span
+        try:
+            message = kwargs.get("message") or args[0]
+            status_code = kwargs.get("status_code")
+            span = async_tracer.active_span
 
-        if all([span, status_code, message]) and (500 <= status_code <= 599):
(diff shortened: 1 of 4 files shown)


instana/instrumentation/sanic_inst.py
```diff
@@ -17,17 +17,21 @@
 
     @wrapt.patch_function_wrapper('sanic.exceptions', 'SanicException.__init__')
     def exception_with_instana(wrapped, instance, args, kwargs):
-        message = kwargs.get("message", args[0])
-        status_code = kwargs.get("status_code")
-        span = async_tracer.active_span
+        try:
+            message = kwargs.get("message") or args[0]
```
andrewslotin, line 21: Would it be more idiomatic to do
```suggestion
            message = kwargs.get("message", args[0])
```
?

andrewslotin (approved):
Overall LGTM, just added a minor suggestion

instana/instrumentation/sanic_inst.py
pdimitra: No, this it the actual bug. This is what I had originally and the issue is that `.get` is a method and it need to have all its arguments evaluated before execution (unfortunately no lazy evaluation ), therefore if you provide message as a keyword argument is going to fire an exception, since it will still attempt to evaluate `arg[0]` which does not exist and it will cause an IndexError

andrewslotin: Fair enough!

PR #340, opened 2021-10-04
limit sanic version in requirements file

pdimitra:
Sanic version 21.9.0 and above is breaking the testing suite, it needs to be investigated what changes we may need to continue supporting newer versions

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -22,7 +22,7 @@ pyramid>=2.0
 pytest>=6.2.4
 pytest-celery
 redis>=3.5.3
-sanic>=19.0.0
+sanic>=19.0.0,<21.9.0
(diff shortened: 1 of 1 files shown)


andrewslotin:
Would it make sense to fix the instrumentation instead of silencing the test?

pdimitra:
I tried to fix it, but it is not that simple, it skips the instrumentation. It could be related to some Signals they introduced, they even print warnings when starting a Sanic app. The documentation has a note that these signals are an experimental feature. Full support will not be finalized until v21.12
So far there is no match info about all these new introductions and it feels to me a waste of time trying to comply with it. I want to make a ticket out if it though and the same applies for all other instrumentations, for which we hold an upper limit in the requirements file.

PR #345, opened 2022-02-06
fix(compose): Fix and modernise compose.yml

Ferenc-:
* Adds a few needed environment variables
* Bumps the versions of `redis` and `mongo` to the same levels used by circleci 
* Enables starting the stack with podman (Without using DNS in Container Networks)
* Minimizes the scope of shared volumes to the absolutely needed config files
* Uses the `:Z` to have the runtime label the content with a private unshared label.
* Removes unused services
* Removes outdated comments

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,50 +1,12 @@
-version: '2'
+version: '3.8'
 services:
   redis:
-    image: redis:4.0.6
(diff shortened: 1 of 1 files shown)


cedricziel (approved):

andriikushch (approved):
LGTM 👍

PR #348, opened 2022-02-21
feat(ci): Add a CI test job for Python 3.10

Ferenc-:
This PR enables the codebase to be tested on Python 3.10 in a CI loop,
so developers get fast feedback of test results on Python 3.10.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -154,6 +154,31 @@ jobs:
             . venv/bin/activate
             pytest -v
 
+  python310:
+    docker:
+      - image: circleci/python:3.10.0-buster
+      - image: circleci/postgres:9.6.5-alpine-ram
+      - image: circleci/mariadb:10-ram
+      - image: circleci/redis:5.0.4
+      - image: rabbitmq:3.5.4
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -154,6 +154,31 @@ jobs:
...
 
+  python310:
+    docker:
+      - image: circleci/python:3.10.0-buster
+      - image: circleci/postgres:9.6.5-alpine-ram
+      - image: circleci/mariadb:10-ram
+      - image: circleci/redis:5.0.4
+      - image: rabbitmq:3.5.4
+      - image: circleci/mongo:4.2.3-ram
+      - image: singularities/pubsub-emulator
```
andrewslotin, line 165: I'd suggest to pin the image version here to make sure that test results are not affected by the upstream changes

andrewslotin (approved):

.circleci/config.yml
Ferenc-: Absolutely true. I was led by the pattern of the previous jobs.
I will pin every image and in general clean up the image mess in a new commit.

PR #352, opened 2022-03-04
Handle X-INSTANA-L according to specifications

Ferenc-:
* Propagate X-INSTANA-L
* Only return spans if X-INSTANA-L==0
* Only propagate span_id and trace_id if  X-INSTANA-L!=0

### DIFF
--- a/instana/propagators/base_propagator.py
+++ b/instana/propagators/base_propagator.py
@@ -138,7 +138,7 @@ class BasePropagator(object):
         """
         This method determines the span context depending on a set of conditions being met
         Detailed description of the conditions can be found here:
-        https://github.com/instana/technical-documentation/tree/master/tracing/specification#http-processing-for-instana-tracers
+        https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#http-processing-for-instana-tracers
         :param trace_id: instana trace id
         :param span_id: instana span id
         :param level: instana level
@@ -157,6 +157,13 @@ class BasePropagator(object):
             correlation = True
 
         ctx_level = self._get_ctx_level(level)
+        if ctx_level == 0 or level == '0':
+            ctx.level = ctx_level
+            trace_id = ctx.trace_id = None
+            span_id = ctx.span_id = None
+            ctx.correlation_type = None
+            ctx.correlation_id = None
+#           return ctx
 
         if trace_id and span_id:
             ctx.trace_id = trace_id[-16:]  # only the last 16 chars
@@ -263,7 +270,7 @@ class BasePropagator(object):
 
     def extract(self, carrier, disable_w3c_trace_context=False):
         """
-        This method overrides the one of the Baseclass as with the introduction of W3C trace context for the HTTP
+        This method overrides one of the Baseclasses as with the introduction of W3C trace context for the HTTP
         requests more extracting steps and logic was required
         :param disable_w3c_trace_context:
         :param carrier:
--- a/instana/propagators/base_propagator.py
+++ b/instana/propagators/base_propagator.py
@@ -137,8 +137,8 @@ class BasePropagator(object):
                                  disable_w3c_trace_context):
         """
         This method determines the span context depending on a set of conditions being met
-        Detailed description of the conditions can be found here:
-        https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#http-processing-for-instana-tracers
+        Detailed description of the conditions can be found in the instana internal technical-documentation,
+        under section http-processing-for-instana-tracers
         :param trace_id: instana trace id
         :param span_id: instana span id
         :param level: instana level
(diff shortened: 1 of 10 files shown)


instana/recorder.py
```diff
@@ -63,7 +63,8 @@ def queued_spans(self):
             except queue.Empty:
                 break
             else:
-                spans.append(span)
+                if span.l != 0:
+                    spans.append(span)
```
andrewslotin, line 67: Would it make sense instead of filtering out suppressed spans on read not to store them at all? With current implementation the queue length returned by `queue_size()` does not match the actual number of elements returned by `queued_spans()`.

andrewslotin (changes requested):
Overall LGTM, however I've put some suggestions to make the changes less error-prone

instana/recorder.py
Ferenc-: Absolutely true. I should fix that.

Ferenc-: Changed.

instana/recorder.py
```diff
@@ -87,8 +87,9 @@ def record_span(self, span):
                 service_name = self.agent.options.service_name
                 json_span = SDKSpan(span, source, service_name)
 
-            # logger.debug("Recorded span: %s", json_span)
-            self.agent.collector.span_queue.put(json_span)
+            if json_span.l != 0:
```
andrewslotin, line 90: I'd suggest we do this at the very beginning and ignore span completely if `level` is set to `0`

Ferenc-: Ooops. Indeed. Fixed.

instana/propagators/http_propagator.py
```diff
@@ -33,35 +34,28 @@ def inject(self, span_context, carrier, disable_w3c_trace_context=False):
                 if traceparent and tracestate:
                     carrier[self.HEADER_KEY_TRACEPARENT] = traceparent
                     carrier[self.HEADER_KEY_TRACESTATE] = tracestate
-                carrier[self.HEADER_KEY_T] = trace_id
-                carrier[self.HEADER_KEY_S] = span_id
-                carrier[self.HEADER_KEY_L] = "1"
+                if level != 0:
```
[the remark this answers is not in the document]
Ferenc-: The condition in the statement can be inverted if we do an early return.
The way I understand the technical specification (technical-documentation/tree/master/tracing/specification#sending-x-instana-l), populating `carrier[self.HEADER_KEY_L]`, is not optional but a `MUST`.
It is possible, that I might have misunderstood something, but not setting it definitely results in huge ammount of UT TC failures for sure, I think those were expecting the previous logic, which was populating that header unconditionally.
So if we consider changing all those existing UT TCs, then let's do it in a different thread.

In the meantime what I think we could do to make the logic more robust on top of the current changes is like this:
```diff
                 if traceparent and tracestate:
                     carrier[self.HEADER_KEY_TRACEPARENT] = traceparent
                     carrier[self.HEADER_KEY_TRACESTATE] = tracestate
-                if level != 0:
-                    carrier[self.HEADER_KEY_T] = trace_id
-                    carrier[self.HEADER_KEY_S] = span_id
+
                 carrier[self.HEADER_KEY_L] = serializable_level
+
+                if level == 0:
+                    return
+
+                carrier[self.HEADER_KEY_T] = trace_id
+                carrier[self.HEADER_KEY_S] = span_id

```
How about this?

instana/propagators/base_propagator.py
```diff
@@ -157,6 +157,12 @@ def __determine_span_context(self, trace_id, span_id, level, synthetic, tracepar
             correlation = True
 
         ctx_level = self._get_ctx_level(level)
+        if ctx_level == 0 or level == '0':
+            ctx.level = ctx_level
```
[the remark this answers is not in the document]
Ferenc-: I have moved the assignment to the very end of the method, so there are no previous calls which could intervene.

instana/propagators/http_propagator.py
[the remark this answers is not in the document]
Ferenc-: I committed the early return for now.

andrewslotin (approved):
LGTM! As a suggestion I believe it might be a good time to refactor the way we populate carrier fields in HTTP propagator to reduce duplication. You might also consider using a boolean value to store the `context.level` for readability.

instana/propagators/base_propagator.py
```diff
@@ -130,15 +130,22 @@ def _get_participating_trace_context(self, span_context):
         traceparent = span_context.traceparent
         tracestate = span_context.tracestate
         traceparent = self._tp.update_traceparent(traceparent, tp_trace_id, span_context.span_id, span_context.level)
+
+        # In suppression mode do not update the tracestate and
+        # do not add the 'in=' key-value pair to the incoming tracestate
+        # Just propagate the incoming tracestate (if any) unchanged.
+        if span_context.level == 0:
```
andrewslotin, line 137: Since we have a helper method introduced now
```suggestion
        if span_context.suppression:
```

andrewslotin (approved):
Overall LGTM.

PR #355, opened 2022-03-13
Fix AttributeError when extracting tags from pika

### LINKED ISSUE #354: AttributeError when extracting tags from pika BlockingChannel
Typo in https://github.com/instana/python-sensor/blob/master/instana/instrumentation/pika.py#L121 causes `AttributeError` when setting span tags. 

```python
2022-03-10 15:47:48,593: 1 DEBUG instana: basic_consume_with_instana:
  File "/usr/local/lib/python3.8/site-packages/instana/instrumentation/pika.py", line 121, in _cb_wrapper
    _extract_consumer_tags(scope.span,
  File "/usr/local/lib/python3.8/site-packages/instana/instrumentation/pika.py", line 33, in _extract_consumer_tags
    _extract_broker_tags(span, conn)
  File "/usr/local/lib/python3.8/site-packages/instana/instrumentation/pika.py", line 21, in _extract_broker_tags
    span.set_tag("address", "%s:%d" % (conn.params.host, conn.params.port))
AttributeError: 'BlockingConnection' object has no attribute 'params'
```

Pika's `BlockingConnection` uses composition instead of inheritance. Therefore, the `params` field is not available in its parent but rather in its `_impl` attribute. This is done for `consume_with_instana` wrapper but not for `basic_consume_with_instana` wrapper.

rlopes-ki:
Resolves #354.

Also removes duplicate call to `span.set_tag` as it seems to already be done at `_extract_broker_tags`.

### DIFF
--- a/instana/instrumentation/pika.py
+++ b/instana/instrumentation/pika.py
@@ -32,7 +32,6 @@ try:
     def _extract_consumer_tags(span, conn, queue):
         _extract_broker_tags(span, conn)
 
-        span.set_tag("address", "%s:%d" % (conn.params.host, conn.params.port))
         span.set_tag("sort", "consume")
         span.set_tag("queue", queue)
 
--- a/instana/instrumentation/pika.py
+++ b/instana/instrumentation/pika.py
@@ -119,7 +119,7 @@ try:
             with tracer.start_active_span("rabbitmq", child_of=parent_span) as scope:
                 try:
                     _extract_consumer_tags(scope.span,
-                                           conn=instance.connection,
+                                           conn=instance.connection._impl,
                                            queue=queue)
                 except:
                     logger.debug("basic_consume_with_instana: ", exc_info=True)
--- a/instana/version.py
+++ b/instana/version.py
@@ -3,4 +3,4 @@
 
 # Module version file.  Used by setup.py and snapshot reporting.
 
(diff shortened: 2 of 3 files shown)


Ferenc-:
Hi,
Thank you for your great contribution!
I noticed, that the main reason why we can have such bugs, is that we have no unit tests,  
which would test with `BlockingConnection`.  
Even our `TestPikaBlockingChannel` class only uses `Connection` and no `BlockingConnection`.  
Do you think you would have the capacity to extend your PR with a change in `tests/clients/test_pika.py`, so it includes coverage for a `BlockingConnection` scenario?

rlopes-ki:
Hi @Ferenc-, 

I don't have much availability right now, and I'm not that comfortable with the `unittest` framework, but I will give it a try.

rlopes-ki:
Sorry for the delay, had to take some time to look into `pika` implementation details. Let me know what you think about these changes.

Ferenc-:
> Sorry for the delay, had to take some time to look into `pika` implementation details. Let me know what you think about these changes.

Looks very good to me.
My only comment would have been the unused class attributes, that you have also fixed in the meantime, before I sent the comment.
So I am merging this now. Many thanks for the contribution.
If you have further improvements then feel free to open new PRs.

PR #359, opened 2022-03-17
chore(github): Add issue template

Ferenc-:
This commit adds an issue template similar to the setup of the `go-sensor` repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -58,10 +58,12 @@ body:
       description: |
         Please the list of environment variables available for the application.
         For example
-        `for pid in $(pidof python3); do
-           echo "#### PID: ${pid} ####";
-           cat /proc/${pid}/environ | tr '\0' '\n';
-         done`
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,67 @@
...
+    id: python-environment
+    attributes:
+      label: Python Environment
+      description: |
+        Please the list of environment variables available for the application.
+        For example
+        `for pid in $(pidof python3); do
+           echo "#### PID: ${pid} ####";
+           cat /proc/${pid}/environ | tr '\0' '\n';
+         done`
```
andrewslotin, line 64: Since you have this code formatted in Markdown, would it make sense to display it the same way in HTML?
```suggestion
        ```
        for pid in $(pidof python3); do
          echo "#### PID: ${pid} ####";
          cat /proc/${pid}/environ | tr '\0' '\n';
        done
        ```
```

andrewslotin (approved):
👍

PR #374, opened 2022-06-23
use ContextVarsScopeManager instead of AsyncioScopeManager for python >= 3.7

BohdanKaminskyi:
Per opentracing [documentation](https://github.com/opentracing/opentracing-python#scope-managers), `ContextVarsScopeManager` should be used for async code in python >= 3.7 . 

For me, main issue was that, `parent_span` context was not propagated for coroutines running inside `asyncio.wait()` or `asyncio.gather()`

### DIFF
--- a/instana/singletons.py
+++ b/instana/singletons.py
@@ -87,13 +87,20 @@ def set_agent(new_agent):
 # this package.
 tracer = InstanaTracer(recorder=span_recorder)
 
-if sys.version_info >= (3, 4):
-    try:
+
+try:
+    if (3, 4) <= sys.version_info < (3, 7):
         from opentracing.scope_managers.asyncio import AsyncioScopeManager
 
         async_tracer = InstanaTracer(scope_manager=AsyncioScopeManager(), recorder=span_recorder)
-    except Exception:
-        logger.debug("Error setting up async_tracer:", exc_info=True)
+
+    elif sys.version_info >= (3, 7):
+        from opentracing.scope_managers.contextvars import ContextVarsScopeManager
+
+        async_tracer = InstanaTracer(scope_manager=ContextVarsScopeManager(), recorder=span_recorder)
+
+except Exception:
+    logger.debug("Error setting up async_tracer:", exc_info=True)
 
 # Mock the tornado tracer until tornado is detected and instrumented first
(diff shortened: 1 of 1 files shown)


BohdanKaminskyi:
Hi @pglombardo , it looks like there are 4 tests which failing due to difference in inner realization of `ContextVarsScopeManager` to `AsyncioScopeManager`: it lacks `scope_manager._set_task_scope` method, instead it uses `_set_scope` method but with different signature. Not sure how to properly fix these tests for py37+ .
Could you please suggest what can I do to fix the issue? 
Thank you!

pglombardo:
Hi @BohdanKaminskyi - I no longer maintain this repository and have lost track of recent opentracing/telemetry scope management changes.  Someone from the team should pick up on this soon I would think.

BohdanKaminskyi:
@Ferenc- Hi, are you maintaining this repo? Maybe you can assist me on issue I mentioned https://github.com/instana/python-sensor/pull/374#issuecomment-1164385117

Ferenc-:
Hi @BohdanKaminskyi,

Thank you for taking the initiative!  
Would it be possible to start by creating a new test case for your original issue,  
so your use case doesn't get broken in a future iteration?

Support for pre `3.7` runtimes is about to be dropped from the `main` branch,  
so all the version checking will likely not be needed at the very end.  

Well about the different signature of `_set_task_scope` vs `_set_scope`,  
namely the second not taking a `task` argument,  
the way I see it, that appears to be the [commented purpose](https://github.com/opentracing/opentracing-python/blob/e2543115c1a9fbb4a49e600f862d54ff4f45f703/opentracing/scope_managers/contextvars.py#L38) of `ContextVarsScopeManager`,  
that you don't have to deal with setting the scope of the task manually,  
but the scope should be propagated automatically.  
While with the old `AsyncioScopeManager` it was [explicitely needed to do manually](https://github.com/opentracing/opentracing-python/blob/e2543115c1a9fbb4a49e600f862d54ff4f45f703/opentracing/scope_managers/asyncio.py#L38).
So is it possible, that we just have to stop trying to do that manually?

Ferenc-:
Not relevant anymore. Closing.

PR #378, opened 2022-07-14
fix(django): Detect running django based on the injected path

Ferenc-:
The old logic relied on markers placed by the decomissioned magic wand tool.
Now we base the detection on a custom path set by the injector.


Ferenc-:
Wrong branch

PR #402, opened 2023-01-10
Handle faulty announce responses from agents

Ferenc-:
* Handle scenarios gracefully, when the agent does not return an expected response
* In particular, avoid the `KeyError` scenario, which was seen in ZD 27830, because of an incomplete response:
````
  File "/usr/local/lib/python3.7/site-packages/instana/agent/host.py", line 150, in set_from
  self.announce_data = AnnounceData(pid=res_data['pid'], agentUuid=res_data['agentUuid'])
  KeyError: 'pid'
````
* Adds test cases for covering these scenarios

### DIFF
--- a/instana/agent/host.py
+++ b/instana/agent/host.py
@@ -123,19 +123,12 @@ class HostAgent(BaseAgent):
 
         return False
 
-    def set_from(self, json_string):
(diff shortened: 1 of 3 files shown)


basti1302 (approved):
LGTM

PR #409, opened 2023-02-15
fix: Handle when the announce response has no fields

Ferenc-:
Handle `AttributeError: 'list' object has no attribute 'get'`, which has been seen in ZD 31421.


pvital (approved):
Looks good.

PR #415, opened 2023-02-21
Upgrade version of tests' dependencies.

pvital:
Most of the upgrades are related to security issues of the packages or their dependencies (transitive).

### DIFF
--- a/tests/requirements-307.txt
+++ b/tests/requirements-307.txt
@@ -1,15 +1,15 @@
 aiofiles>=0.5.0
 aiohttp>=3.8.3
 boto3>=1.17.74
+
+# TODO: importlib_metadata package removed deprecated entry_point interfaces on 
+# version 5.0.0 [1], and this change impacts celery >= 5.2.7 [2] running with 
+# python 3.7.X (it doesn't impact >= 3.8). For this reason, we control celery 
+# and importlib_metadata versions on python 3.7 environments.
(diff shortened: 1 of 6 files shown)


tests/requirements-307.txt
```diff
@@ -1,26 +1,26 @@
...
-# the release notes: https://github.com/celery/kombu/releases/tag/v5.3.0b2
-kombu>=5.3.0b2
+
+# TODO: importlib_metadata package removed deprecated entry_point interfaces on 
+# version 5.0.0 [1], and this change impacts celery >= 5.2.7 [2] running with 
+# python 3.7.X (it doesn't impact >= 3.8). For this reason, we control celery 
+# and importlib_metadata versions on python 3.7 environments.
+# [1] https://github.com/python/importlib_metadata/pull/405
+# [2] https://github.com/celery/celery/issues/7783
+celery>=4.3.0,<=5.2.7
```
Ferenc-, line 11: I think this line could be `celery>=5.2.7` like in the rest of the requirements.

tests/requirements-307.txt
```diff
@@ -1,26 +1,26 @@
...
+importlib-metadata<=4.13.0
```
Ferenc-, line 12: I think the recommended is `importlib-metadata<5.0.0`. Although unlikely to be release, but let's say there is a theoretical `4.13.1` security fix, and that should still be tested in this mix.

PR #416, opened 2023-03-06
feat: improve robustness of custom service naming

basti1302:
Previously, we only added span.data.service to _entry_ spans (when the environment variable INSTANA_SERVICE_NAME is set). The requirements around this have changed, now we add this annotation to _all_ spans (but still only if it has been explicitly configured).

See also:
* https://github.ibm.com/instana/technical-documentation/pull/290/files
* https://instana.slack.com/archives/GBS2G15GD/p1675160462737709


Ferenc-:
LGTM

PR #421, opened 2023-05-15
test[fix]: Pass parameters to aiohttp correctly

Ferenc-:
Apparently the parameter passing in our `aiohttp` tests were incorrect, and was depending on a `yarl` bug [1] that was fixed [2] in `1.9.0`, and with this new yarl version, our tests started to fail. This fix corrects the parameter passing, so we can use the latest `yarl`.

[1] https://github.com/aio-libs/yarl/issues/723
[2] https://github.com/aio-libs/yarl/pull/792


pvital (approved):
LGTM

PR #423, opened 2023-05-16
fix(blinker): Always set 500 on exception signal

Ferenc-:
Removes a `TODO` item, and re-enables the assertion of the status code being `500` on the `wsgi` span.


pvital (approved):
LGTM

PR #426, opened 2023-05-19
fix(w3c trace context): consider traceparent format version ff invalid

basti1302:
This is a follow up to commit
f283a4beb4d2b7ccb9b5aae55d076c7a5156e1d0
which added some tests for parsing newer versions. One of the tests used format version "ff", which the W3C trace context specification explicitly defines as invalid. This commit fixes that test and also the regex used for parsing the traceparent header, so that a header with version ff will be rejected as invalid.

### DIFF
--- a/instana/w3c_trace_context/traceparent.py
+++ b/instana/w3c_trace_context/traceparent.py
@@ -9,7 +9,7 @@ SAMPLED_BITMASK = 0b1;
 
 class Traceparent:
     SPECIFICATION_VERSION = "00"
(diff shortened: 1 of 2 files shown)


basti1302:
Sorry @instana/eng-python to bug you again with W3C trace context stuff. I temporarily had forgotten that version ff is explicitly forbidden by the spec and unfortunately used exactly that invalid version in the test I added a few days ago :-/

Ferenc- (approved):
Approved

PR #427, opened 2023-05-22
Fix: remove the check for the Server header of the Instana agent.

pvital:
The check for the Server header of the Instana agent can break the announce procedure in service meshes when there is a proxy between the tracer and the Instana agent. This happens because some proxies do not forward the original Server header. So instead, the collector checks whether the HTTP status code is in the 2xx range.

### DIFF
--- a/instana/agent/host.py
+++ b/instana/agent/host.py
@@ -41,7 +41,6 @@ class HostAgent(BaseAgent):
     """
     AGENT_DISCOVERY_PATH = "com.instana.plugin.python.discovery"
     AGENT_DATA_PATH = "com.instana.plugin.python.%d"
-    AGENT_HEADER = "Instana Agent"
 
     def __init__(self):
         super(HostAgent, self).__init__()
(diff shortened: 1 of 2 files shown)


instana/agent/host.py
```diff
@@ -157,15 +156,16 @@ def is_agent_listening(self, host, port):
         result = False
         try:
             url = "http://%s:%s/" % (host, port)
-            response = self.client.get(url, timeout=0.8)
+            response = self.client.get(url, timeout=5000)
```
basti1302, line 159: _issue:_ If I read https://requests.readthedocs.io/en/latest/user/advanced/#timeouts correctly, the timeout is specified as the number of seconds, not milliseconds. That is probably why it was 0.8 before? 5000 _seconds_ would be a really long timeout. 5 seconds would be good I guess.

basti1302 (commented):
Looks good. We should double check the timeout change, see inline comment.

Ferenc- (changes requested):
Good work so far. We are getting there.
The [coverage report shows](https://output.circle-artifacts.com/output/job/27c6f31a-d745-446b-81cf-c91f480b88eb/artifacts/0/htmlcov/d_8932aa06527cfb99_host_py.html#t161) that the failed attempt is nicely covered with the test case,
while the successful attempt is not yet covered.

Although the production code around the succesful attempt is minuscule.
I still think covering the succesful attempt would be valuable, not just because we can't have a functioning system without a successful connection attempts,
but also because this change is about enabling successful connections in a broader situation.
so any other change is secondary priority here.

tests/platforms/test_host.py
```diff
@@ -10,6 +10,7 @@
 from mock import MagicMock, patch
 
 import requests
+import requests_mock
```
Ferenc-, line 13: Where do we use this?

Ferenc- (approved):
Approved

PR #429, opened 2023-06-02
ci: Post a message on Slack when a new release is published

Ferenc-:


bin/post_release_on_slack.py
pvital, line 1: Not a blocker, but WDYT about changing the filename to `announce_release_on_slack.py`?

pvital (approved):
LGTM

pvital (approved):

basti1302 (approved):

PR #430, opened 2023-06-12
chore: Pre-release version for testing with tracer TS

Ferenc-:


CagriYonca:
Closing this PR since we're already supporting newer versions of instana.

PR #434, opened 2023-07-07
feat: Add new lambda regions

Ferenc-:


bin/aws-lambda/build_and_publish_lambda_layer.py
```diff
@@ -71,9 +71,35 @@
...
+               'eu-west-1',
+               'eu-west-2',
+               'eu-west-3',
+               'me-central-1',
+               'me-south-1',
+               'sa-east-1',
+               'us-east-1',
+               'us-east-2',
+               'us-west-1',
+               'us-west-2'
```
basti1302, line 101: @Ferenc- 

This still is missing a few regions. If I am not mistaken you have 27 regions, while it should be 30.

These regions are missing in your list I think: af-south-1, ap-east-1, il-central-1

Here is the full list of regions we should publish to:

```
af-south-1
ap-east-1
ap-northeast-1
ap-northeast-2
ap-northeast-3
ap-south-1
ap-south-2
ap-southeast-1
ap-southeast-2
ap-southeast-3
ap-southeast-4
ca-central-1
cn-north-1
cn-northwest-1
eu-central-1
eu-central-2
eu-north-1
eu-south-1
eu-south-2
eu-west-1
eu-west-2
eu-west-3
il-central-1
me-central-1
me-south-1
sa-east-1
us-east-1
us-east-2
us-west-1
us-west-2
```

basti1302 (changes requested):

bin/aws-lambda/build_and_publish_lambda_layer.py
Ferenc-: Thank you for pointing it out. I have filled up those regions as well, and made a change here.

PR #446, opened 2023-08-24
fix(pep0249): Use the wrapped value in CTX instead of unwrapped

Ferenc-:


pvital (approved):
LGTM

I tested it in two different ways, with a Flask app connecting to a PostgreSQL server. The first test was using the built-in Flask web server [1], and the second one was with uWSGI [2]. Both tests used the following code, and they worked, as you can see in [1] and [2].

```python
    with get_db_connection() as conn:
        with conn.cursor() as cur:
            cur.execute(
                'SELECT * from Users;'
            )
            users = cur.fetchall()
```

[1] https://instana.rocks/s/-h9kZcw7RqunpU_WibydYQ
[2] https://instana.rocks/s/hfmukSbUSDqJ3t1kfUNCRg

GSVarsha (commented):
```python
with psycopg2.connect(**dsn) as connection:

    with connection.cursor() as cursor:
        cursor.execute(f"select * from {table}")
        employees = cursor.fetchall()
  
    with connection.cursor() as cursor:
        cursor.execute(f"select * from {table} where id=1")
        emp1 = cursor.fetchall()
```
Verified the changes using the below code with 
1. `gunicorn`

```shell
gunicorn -w 4 --bind 0.0.0.0:8000 app:app
``` 

2. Flask's default web server

```shell
python -m flask --debug run
```

PR #452, opened 2023-09-01
refactor(tests): Clean up old PEP-249 tests

Ferenc-:
* Use the `unittest` module's `assert` in unit tests.
* Drop legacy branching for unsupported 2.7
* Eliminate unused variables
* Remove superfluous initializations
* Use unpack assignments

### DIFF
--- a/tests/clients/test_mysqlclient.py
+++ b/tests/clients/test_mysqlclient.py
@@ -11,41 +11,7 @@ from ..helpers import testenv
 from unittest import SkipTest
 from instana.singletons import tracer
 
-
 logger = logging.getLogger(__name__)
-
-create_table_query = 'CREATE TABLE IF NOT EXISTS users(id serial primary key, \
-                      name varchar(40) NOT NULL, email varchar(40) NOT NULL)'
-
-create_proc_query = """
-CREATE PROCEDURE test_proc(IN t VARCHAR(255))
-BEGIN
-    SELECT name FROM users WHERE name = t;
-END
-"""
-
-db = MySQLdb.connect(host=testenv['mysql_host'], port=testenv['mysql_port'],
-                     user=testenv['mysql_user'], passwd=testenv['mysql_pw'],
-                     db=testenv['mysql_db'])
-
-cursor = db.cursor()
-cursor.execute(create_table_query)
-
-while cursor.nextset() is not None:
-    pass
-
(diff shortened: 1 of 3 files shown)


pvital (commented):
Great work! We should review all the tests and refactor them. 

I have one comment only: it's not a big deal, but it's a best practice for readability.

tests/clients/test_mysqlclient.py
```diff
@@ -11,10 +11,7 @@
 from instana.singletons import tracer
 
 
-if sys.version_info[0] > 2:
-    import MySQLdb
-else:
-    raise SkipTest("mysqlclient supported on Python 3 only")
+import MySQLdb
```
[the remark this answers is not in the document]
Ferenc-: Done.

tests/clients/test_mysqlclient.py
```diff
@@ -75,24 +71,22 @@ def test_vanilla_query(self):
         self.assertEqual(0, len(spans))
 
     def test_basic_query(self):
-        result = None
         with tracer.start_active_span('test'):
             result = self.cursor.execute("""SELECT * from users""")
             self.cursor.fetchone()
```
GSVarsha, line 76: If we just assert on the number of rows returned (i.e. result), we can remove the `cursor.fetch*` statements right?

Ferenc-: Good catch! This comment has revealed an even bigger issue.
Namely that the number of the rows in the database are completely out of control.
For example if you keep running the tests `test_mysqlclient.py`, then the number of rows keep increasing, because there is a TC which keeps inserting, and nothing deletes...
Also because the order in which the TCs are running are undefined, as they should not impact each other (but they do), currently the resul in ` result = self.cursor.execute("""SELECT * from users""")` is also undeterministic.
That is also the reason why previously there was no definite value `test_basic_query` like `result >= 0`, because it was impossible to tell with certainty. Now all that is fixed up in a new commit. Please check

GSVarsha: @Ferenc- 
The changes look perfect! Can we make similar changes to the other 2 files (test_pymysql.py, test_psycopg2.py) just to be consistent?

Ferenc-: Right. Now I have fixed those too.

GSVarsha (approved):
Looks Good To Me!

PR #460, opened 2023-10-17
Fork missing spans fix

pvital:
In multithreaded applications (services), only the first or master process is announced to the Agent when the service starts, but if a not announced forked process handles the (HTTP) ENTRY requests and it contains one or more EXIT spans created during the first request, those EXIT spans are missed and not reported. 

This PR has a fix that returns `True` when the `agent.host.can_send()` method is called and detects the state machine is in `wait4init` or `good2go` states, which happens after the current process is announced to the Agent.

### DIFF
--- a/instana/agent/host.py
+++ b/instana/agent/host.py
@@ -115,7 +115,6 @@ class HostAgent(BaseAgent):
             self._boot_pid = current_pid
             logger.debug("Fork detected; Handling like a pro...")
             self.handle_fork()
-            return False
 
         if self.machine.fsm.current in ["wait4init", "good2go"]:
             return True
--- a/instana/collector/base.py
+++ b/instana/collector/base.py
@@ -35,7 +35,7 @@ class BaseCollector(object):
         if env_is_test:
             # Override span queue with a multiprocessing version
             # The test suite runs background applications - some in background threads,
-            # others in background processes.  This multiprocess queue allows us to collect
+            # others in background processes.  This multiprocessing queue allows us to collect
             # up spans from all sources.
             import multiprocessing
             self.span_queue = multiprocessing.Queue()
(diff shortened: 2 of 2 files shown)


instana/instrumentation/pep0249.py
```diff
@@ -28,6 +28,8 @@ def _collect_kvs(self, span, sql):
                 span.set_tag(ext.DATABASE_INSTANCE, self._connect_params[1]['db'])
             elif 'database' in self._connect_params[1]:
                 span.set_tag(ext.DATABASE_INSTANCE, self._connect_params[1]['database'])
+            elif 'dbname' in self._connect_params[1]:
+                span.set_tag(ext.DATABASE_INSTANCE, self._connect_params[1]['dbname'])
```
Ferenc-, line 32:
```
  db_parameter_name = next(p for p in ('db', 'database', 'dbname') if p in self_connect_params[1], None)
  if db_parameter_name:
    span.set_tag(ext.DATABASE_INSTANCE, self._connect_params[1][db_parameter_name])
```

instana/agent/host.py
```diff
@@ -115,7 +115,7 @@ def can_send(self):
             self._boot_pid = current_pid
             logger.debug("Fork detected; Handling like a pro...")
             self.handle_fork()
-            return False
+            return True
```
Ferenc-, line 118: The expected behavior of the `can_send` method has been well summarized on line `105`.
Now on this line `118`, that behavior seems to have been completely destroyed, because this claims that the tracer is in a state in which it can send, but there doesn't seem to be a single check on the state machine, furthermore the call on line 117 even resets the state machine.
A purpose of this `can_send` was to give the recorder a green light for recording spans,
without the high risk of quickly building up a huge queue of spans (1000 or more), and introducing latency in tracing or even worse running out of memory (ending up OOM killed). Now the tracer seems to violate [the specification](https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#span-buffering), which says that `Tracers SHOULD drop spans if they cannot be sent to the agent after collecting 1000 spans or more.` Because previously there was no buffering at all, but now this adds potentially unlimited buffering. because there is nothing else to check and limit the queue_size.

pvital: Got it. I will work on a better solution here then.

Ferenc- (approved):
Approved for now. Let's wait with a new release until we have tracer test suite feedback.

PR #461, opened 2023-10-18
Add `dbname` as collecting parameter for PEP0249

pvital:
This PR contains two commits.

The first one, https://github.com/instana/python-sensor/commit/55e826e81ff311cefac112cd0d1fa4b111c91f23, adds `dbname` as a collecting parameter for PEP0249. When connecting to a PostgreSQL server, for example, using the psycopg2 module, the parameter `dbname` can be used to specify the database name to connect into, like the following piece of code:

```python
    conn = psycopg2.connect(
        host=os.environ['POSTGRES_HOST'],
        dbname=os.environ['POSTGRES_DB'],
        user=os.environ['POSTGRES_USER'],
        password=os.environ['POSTGRES_PASSWORD']
    )
```

Without this fix, the Instana UI is not able to display the database name when inspecting an EXIT span, as shown in the following image:

before_commit_0f9434e02013

With this fix, the name of the database is collected and displayed:

after_commit_0f9434e02013

The second commit, 221432e, adapts all the unit tests that use PEP0249 to guarantee that all options for database parameter names during the connection are tested.

### DIFF
--- a/tests/clients/test_psycopg2.py
+++ b/tests/clients/test_psycopg2.py
@@ -17,9 +17,15 @@ logger = logging.getLogger(__name__)
 
 class TestPsycoPG2(unittest.TestCase):
     def setUp(self):
-        self.db = psycopg2.connect(host=testenv['postgresql_host'], port=testenv['postgresql_port'],
-                                   user=testenv['postgresql_user'], password=testenv['postgresql_pw'],
-                                   database=testenv['postgresql_db'])
+        deprecated_param_name = self.shortDescription() == 'test_deprecated_parameter_database'
+        kwargs = {
+            'host': testenv['postgresql_host'],
+            'port': testenv['postgresql_port'],
+            'user': testenv['postgresql_user'],
+            'password': testenv['postgresql_pw'],
+            'dbname' if not deprecated_param_name else 'database': testenv['postgresql_db'],
+        }
+        self.db = psycopg2.connect(**kwargs)
 
         database_setup_query = """
         DROP TABLE IF EXISTS users;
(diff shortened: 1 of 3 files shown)


tests/clients/test_psycopg2.py
```diff
@@ -19,7 +19,7 @@ class TestPsycoPG2(unittest.TestCase):
     def setUp(self):
         self.db = psycopg2.connect(host=testenv['postgresql_host'], port=testenv['postgresql_port'],
                                    user=testenv['postgresql_user'], password=testenv['postgresql_pw'],
-                                   database=testenv['postgresql_db'])
+                                   dbname=testenv['postgresql_db'])
```
Ferenc-, line 22: Now the scenario in which `db_parameter_name` is `database` and not `dbname` is not covered anymore, when it previously was. We should write variants to `test_basic_query` where we set up connections, with `dbname` `database` and `db` parameters as well.

Ferenc- (changes requested):

tests/clients/test_psycopg2.py
pvital: Well, I can do it, but as described in the `psycopg` [documentation](https://www.psycopg.org/docs/module.html#psycopg2.connect), `database` is a deprecated alias, and `dbname` is the connection parameter to the database name.

tests/clients/test_pymysql.py
```diff
@@ -17,7 +17,7 @@ class TestPyMySQL(unittest.TestCase):
     def setUp(self):
         self.db = pymysql.connect(host=testenv['mysql_host'], port=testenv['mysql_port'],
                                   user=testenv['mysql_user'], passwd=testenv['mysql_pw'],
-                                  db=testenv['mysql_db'])
+                                  database=testenv['mysql_db'])
```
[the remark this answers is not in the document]
pvital: Same as in the `psycopg`, the `pymsql` [documentation](https://pymysql.readthedocs.io/en/latest/modules/connections.html) mentions that `db` is a deprecated alias, and `database` is the connection parameter to the database name.

PR #462, opened 2023-10-20
Fix autotrace detection

Ferenc-:


instana/collector/helpers/runtime.py
```diff
@@ -181,7 +181,7 @@ def _collect_runtime_snapshot(self, plugin_data):
 
             if 'AUTOWRAPT_BOOTSTRAP' in os.environ:
                 snapshot_payload['m'] = 'Autowrapt'
-            elif 'INSTANA_MAGIC' in os.environ:
+            elif '/tmp/.instana/python' in sys.path:
```
pvital, line 184: There are better practices than having hardcoded paths or file names, IMHO. It would be better to have this path as a global configuration variable, and then we can set it once to be used across the code.

In addition, I also found the `INSTANA_MAGIC` variable in the `instana/instrumentation/fastapi_inst.py` file, so I guess that can also be removed.

tests/platforms/test_host_collector.py
```diff
@@ -45,6 +47,8 @@ def tearDown(self):
 
         set_agent(self.original_agent)
         set_tracer(self.original_tracer)
+        if '/tmp/.instana/python' in sys.path:
```
pvital, line 50: As mentioned before, if we have a global configuration variable for this path/file name, it will be easier to maintain the code in the future.

tests/platforms/test_host_collector.py
```diff
@@ -184,6 +194,54 @@ def test_prepare_payload_with_snapshot_disabled_python_packages(self, mock_shoul
...
+        self.assertIn('fysom', snapshot['versions'])
+        self.assertIn('opentracing', snapshot['versions'])
+        self.assertIn('basictracer', snapshot['versions'])
+
+
+    @patch.object(HostCollector, "should_send_snapshot_data")
+    def test_prepare_payload_with_autotrace(self, mock_should_send_snapshot_data):
+        mock_should_send_snapshot_data.return_value = True
+
+        sys.path.append('/tmp/.instana/python')
```
pvital, line 230: Same here, as mentioned before, regarding this hardcoded path/file name.

pvital (changes requested):

tests/platforms/test_host_collector.py
```diff
@@ -184,6 +194,54 @@ def test_prepare_payload_with_snapshot_disabled_python_packages(self, mock_shoul
...
+    def test_prepare_payload_with_autowrapt(self, mock_should_send_snapshot_data):
+        mock_should_send_snapshot_data.return_value = True
+        os.environ["AUTOWRAPT_BOOTSTRAP"] = "instana"
+        self.create_agent_and_setup_tracer()
+
+        payload = self.agent.collector.prepare_payload()
+        self.assertTrue(payload)
+        self.assertIn('snapshot', payload['metrics']['plugins'][0]['data'])
+        snapshot =  payload['metrics']['plugins'][0]['data']['snapshot']
+        self.assertTrue(snapshot)
```
GSVarsha, line 214: Can we have just one assertion here like this?

```suggestion
        snapshot = payload['metrics']['plugins'][0]['data'].get('snapshot')
        self.assertIsNotNone(snapshot)
```

tests/platforms/test_host_collector.py
```diff
@@ -184,6 +194,54 @@ def test_prepare_payload_with_snapshot_disabled_python_packages(self, mock_shoul
...
+
+        sys.path.append('/tmp/.instana/python')
+
+        self.create_agent_and_setup_tracer()
+
+        payload = self.agent.collector.prepare_payload()
+        self.assertTrue(payload)
+        self.assertIn('snapshot', payload['metrics']['plugins'][0]['data'])
+        snapshot =  payload['metrics']['plugins'][0]['data']['snapshot']
+        self.assertTrue(snapshot)
```
GSVarsha, line 238: Same as [previous suggestion](https://github.com/instana/python-sensor/pull/462/files#r1368396729)

tests/platforms/test_host_collector.py
```diff
@@ -184,6 +194,54 @@ def test_prepare_payload_with_snapshot_disabled_python_packages(self, mock_shoul
...
+        self.assertIn('m', snapshot)
+        self.assertEqual('AutoTrace', snapshot['m'])
+        self.assertIn('version', snapshot)
+        self.assertGreater(len(snapshot['versions']), 5)
+        self.assertEqual(snapshot['versions']['instana'], VERSION)
+        self.assertIn('wrapt', snapshot['versions'])
+        self.assertIn('fysom', snapshot['versions'])
+        self.assertIn('opentracing', snapshot['versions'])
+        self.assertIn('basictracer', snapshot['versions'])
```
GSVarsha, line 247: Should we loop through the expected libraries like this? (with or without an optional message when it fails)
```suggestion
        expected_packages = ['wrapt', 'fysom', 'opentracing', 'basictracer']
        for package in expected_packages:
            self.assertIn(package, snapshot['versions'], f"{package} not found in snapshot['versions']")

```

instana/collector/helpers/runtime.py
Ferenc-: Valid point, the hardcoding should be eliminated.

tests/platforms/test_host_collector.py
Ferenc-: The hardcoding will be removed.

tests/platforms/test_host_collector.py
Ferenc-: The hardcoding will be removed.

tests/platforms/test_host_collector.py
Ferenc-: While it is indeed one line shorter, and often the shorter the better, this case is trickier.
When the `assertIn` fails, the test framework immediately and clearly communicates to the developer, what is missing from which collection, without having to check the test code.
When the `assertIsNotNone` fails, then it doesn't have a clear meaning without actually checking the test code.
It doesn't show where it is missing from, furthermore it's not even entirely safe to assume that it is missing, theoretically it could also be that it was set to the `None` value.

tests/platforms/test_host_collector.py
Ferenc-: Same as above.

tests/platforms/test_host_collector.py
Ferenc-: Valid point, we should have a loop.

Ferenc-: done

instana/collector/helpers/runtime.py
```diff
@@ -181,7 +182,7 @@ def _collect_runtime_snapshot(self, plugin_data):
 
             if 'AUTOWRAPT_BOOTSTRAP' in os.environ:
                 snapshot_payload['m'] = 'Autowrapt'
-            elif 'INSTANA_MAGIC' in os.environ:
```
pvital, line 184: Is the `INSTANA_MAGIC` variable in the `instana/instrumentation/fastapi_inst.py` still needed?

Ferenc-: That is a different can of worms that belongs to a separate PR.

pvital (approved):
LGTM

PR #469, opened 2023-11-06
bin: Expect separate profiles for CN region

Ferenc-:


bin/aws-lambda/build_and_publish_lambda_layer.py
```diff
@@ -107,14 +121,17 @@
...
     print("===> Uploading layer to AWS %s " % region)
+    profile = 'china' if region in cn_regions else 'non-china'
+
     response = check_output(["aws", "--region", region, "lambda", "publish-layer-version",
                              "--description",
                              "Provides Instana tracing and monitoring of AWS Lambda functions built with Python",
                              "--license-info", "MIT", "--output", "json",
                              "--layer-name", LAYER_NAME, "--zip-file", aws_zip_filename,
-                             "--compatible-runtimes", "python3.7", "python3.8", "python3.9", "python3.10"])
+                             "--compatible-runtimes", "python3.7", "python3.8", "python3.9", "python3.10",
```
pvital, line 133: Taking advantage of the PR, removing the `python3.10` entry as compatible runtime would be good since we still don't support this version officially, and the [documentation](https://www.ibm.com/docs/en/instana-observability/current?topic=lambda-aws-native-tracing-python#supported-runtimes) doesn't list it. This "issue" was raised by a customer [here](https://instana.slack.com/archives/C2GAQMQLC/p1698241371621549).

```suggestion
                             "--compatible-runtimes", "python3.7", "python3.8", "python3.9",
```

pvital (dismissed):
Just an extra change request. But it looks good.

bin/aws-lambda/build_and_publish_lambda_layer.py
Ferenc-: OK, that will be the topic of https://github.com/instana/python-sensor/pull/470, because our interest would be to be able to quickly revert such a change.

PR #475, opened 2023-11-28
refactor: Using context manager to handle threading.Lock

pvital:
This PR solves the issue of the background report thread not releasing the lock when the application is manually instrumented and runs with multiple workers (uWSGI).

### DIFF
--- a/instana/collector/base.py
+++ b/instana/collector/base.py
@@ -150,16 +150,11 @@ class BaseCollector(object):
         Prepare and report the data payload.
         @return: Boolean
         """
-        if env_is_test is False:
-            lock_acquired = self.background_report_lock.acquire(False)
-            if lock_acquired:
-                try:
-                    payload = self.prepare_payload()
-                    self.agent.report_data_payload(payload)
-                finally:
-                    self.background_report_lock.release()
-            else:
-                logger.debug("prepare_and_report_data: Couldn't acquire lock")
+        if env_is_test:
+            return True
+        with self.background_report_lock:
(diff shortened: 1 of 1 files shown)


instana/collector/base.py
```diff
@@ -151,15 +151,12 @@ def prepare_and_report_data(self):
         @return: Boolean
         """
         if env_is_test is False:
```
Ferenc-, line 153: Since we are refactoring, I wolud refactor this statement to an early return so the rest of the logic (beginning with a `with` statement) is not behind a condition.

instana/collector/base.py
```diff
@@ -151,15 +151,12 @@ def prepare_and_report_data(self):
...
+            with self.background_report_lock:
                 try:
                     payload = self.prepare_payload()
                     self.agent.report_data_payload(payload)
-                finally:
-                    self.background_report_lock.release()
-            else:
-                logger.debug("prepare_and_report_data: Couldn't acquire lock")
+                except:
+                    logger.debug("prepare_and_report_data: Couldn't acquire lock")
```
Ferenc-, line 159: I think the `Couldn't acquire lock` part is obsolete here.

Ferenc- (changes requested):

instana/collector/base.py
pvital: +1

Ferenc- (approved):
LGTM

GSVarsha (approved):
Looks good to me!

PR #478, opened 2023-12-06
ci: Use only the released (non --pre) versions of Django 5 for testing

Ferenc-:


pvital (commented):
It looks good for me.

PR #479, opened 2023-12-07
Use latest urllib3 without secure extra

GSVarsha:
Signed-off-by: Varsha GS varsha.gs@ibm.com

- Use urllib3>=1.26.18
- Remove secure extra feature ([more info](https://github.com/urllib3/urllib3/issues/2680))

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(name='instana',
                         'protobuf<5.0.0',
                         'requests>=2.6.0',
                         'six>=1.12.0',
-                        'urllib3>=1.26.18',],
+                        'urllib3>=1.26.5',],
       entry_points={
                     'instana':  ['string = instana:load'],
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -48,7 +48,7 @@
                         'protobuf<5.0.0',
                         'requests>=2.6.0',
                         'six>=1.12.0',
-                        'urllib3<1.27,>=1.26.5',],
+                        'urllib3>=1.26.18',],
```
Ferenc-, line 51: The way I see there is no reason why customers who had to stick with `urllib3`  version `1.26.5`, should not be able to use our tracer without any issue. So we perhaps we could keep this on `>=1.26.5`?

GSVarsha: Makes sense 👍🏼

pvital (approved):
It looks good to me.

Ferenc- (approved):
Approved

PR #480, opened 2023-12-08
fix(lambda): Record the statusCode when available

Ferenc-:


instana/instrumentation/aws/lambda_inst.py
```diff
@@ -35,6 +36,8 @@ def lambda_handler_with_instana(wrapped, instance, args, kwargs):
                     result['headers']['Server-Timing'] = server_timing_value
                 elif 'multiValueHeaders' in result:
                     result['multiValueHeaders']['Server-Timing'] = [server_timing_value]
+                if 'statusCode' in result:
+                    scope.span.set_tag(ext.HTTP_STATUS_CODE, int(result['statusCode']))
```
basti1302, line 40: _question:_ Do you maybe also need to add the usual "if statusCode >= 500 then span.ec = 1" logic here? (Or is that handled in a central place, like some post-processing for spans?) In any case, adding a test case with status code >= 500 and verifying span.ec=1 might be nice?

basti1302 (approved):

instana/instrumentation/aws/lambda_inst.py
Ferenc-: True. Apparently it wasn't performed in any central place. It's done now.

pvital (approved):
It looks good to me.

PR #482, opened 2023-12-12
fix: The spec doesn't set upper limit for ec=1 if code>=500

Ferenc-:
Quote from the [spec](
https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#capturing-errors-on-http-spans):

> For HTTP exit and entry spans alike, if a full request/response cycle has happened (which implies that an HTTP status code is present): If the HTTP status code is >= 500, the associated span MUST be marked as an error via span.ec.

This means that there is no upper limit, and this should apply to custom status codes as well, which might be above `511`, or even `599`.

### DIFF
--- a/instana/instrumentation/aiohttp/client.py
+++ b/instana/instrumentation/aiohttp/client.py
@@ -51,7 +51,7 @@ try:
                         if custom_header in params.response.headers:
                             scope.span.set_tag("http.header.%s" % custom_header, params.response.headers[custom_header])
 
-                if 500 <= params.response.status <= 599:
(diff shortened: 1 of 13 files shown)


tamarastanojevic (approved):
Right, anything greater or equal than 500 must be considered as error. LGTM 👍🏻

basti1302 (approved):

pvital (approved):
It looks good to me.

GSVarsha (approved):
LGTM!

PR #483, opened 2023-12-13
Test infra updates

pvital:
This PR contains updates and changes to improve and synchronize the test infrastructure.


GSVarsha:
Hey Paulo, are the following intentional?

- We are using [specific version of services](https://github.com/instana/python-sensor/pull/483/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R146) in the `config.yml` file and [the latest available version of services](https://github.com/instana/python-sensor/pull/483/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R43) in the `docker-compose.yml` file.
- The `0` in `passw0rd`

Ferenc- (approved):
OK

Ferenc-:
> * We are using [specific version of services](https://github.com/instana/python-sensor/pull/483/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R146) in the `config.yml` file and [the latest available version of services](https://github.com/instana/python-sensor/pull/483/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R43) in the `docker-compose.yml` file.

It's not just the version which is different, they are entirely different images, `cimg/postgres` vs `postgres`. They contain different things and have different lifecycles. See https://circleci.com/docs/circleci-images/

pvital:
@GSVarsha, yes, all of the changes are intentional :-D 

> We are using [specific version of services](https://github.com/instana/python-sensor/pull/483/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R146) in the config.yml file and [the latest available version of services](https://github.com/instana/python-sensor/pull/483/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R43) in the docker-compose.yml file.

So, the `docker-compose.yml` file is intended to support the services for the local tests during the development using `pytest`. As @Ferenc- has mentioned, the images from the CircleCI config and this docker-compose file differ. Still, they could be set to use the latest versions in both files since we are not dependent, or interested, on the server versioning but on the clients.

> The 0 in passw0rd

Always use numbers (and symbols) in your dummy test passwords. ;-)

GSVarsha (approved):
Thank you Ferenc and Paulo for the explanation.
Looks good to me then!

PR #488, opened 2023-12-27
FastAPI: capture responseHeadersOnEntrySpans

GSVarsha:
FastAPI: capture responseHeadersOnEntrySpans

### DIFF
--- a/instana/instrumentation/asgi.py
+++ b/instana/instrumentation/asgi.py
@@ -20,13 +20,14 @@ class InstanaASGIMiddleware:
         self.app = app
 
     def _extract_custom_headers(self, span, headers):
+        if agent.options.extra_http_headers is None:
+            return        
         try:
-            if agent.options.extra_http_headers is not None:
-                for custom_header in agent.options.extra_http_headers:
-                    # Headers are in the following format: b'x-header-1'
-                    for header_pair in headers:
-                        if header_pair[0].decode('utf-8').lower() == custom_header.lower():
-                            span.set_tag("http.header.%s" % custom_header, header_pair[1].decode('utf-8'))
+            for custom_header in agent.options.extra_http_headers:
+                # Headers are in the following format: b'x-header-1'
+                for header_pair in headers:
+                    if header_pair[0].decode('utf-8').lower() == custom_header.lower():
+                        span.set_tag("http.header.%s" % custom_header, header_pair[1].decode('utf-8'))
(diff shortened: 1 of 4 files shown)


instana/instrumentation/asgi.py
```diff
@@ -21,11 +21,12 @@ def __init__(self, app):
 
     def _extract_custom_headers(self, span, headers):
         try:
-            for custom_header in agent.options.extra_http_headers:
-                # Headers are in the following format: b'x-header-1'
-                for header_pair in headers:
-                    if header_pair[0].decode('utf-8').lower() == custom_header.lower():
-                        span.set_tag("http.header.%s" % custom_header, header_pair[1].decode('utf-8'))
+            if agent.options.extra_http_headers is not None:
```
Ferenc-, line 24: This check should be inverted for an early return and should go outside of the try block.

tests/apps/fastapi_app/app.py
```diff
@@ -24,6 +24,11 @@ async def root():
 async def user(user_id):
     return {"user": user_id}
 
+@fastapi_server.get("/response_headers")
+async def response_headers():
+    headers = {'X-Capture-This-Too': 'this too'}
```
Ferenc-, line 29: I think it is safer to test with multiple headers at a time.

tests/frameworks/test_fastapi.py
```diff
@@ -366,3 +366,53 @@ def test_custom_header_capture(server):
...
+
+
+def test_response_header_capture(server):
+
+     # The background FastAPI server is pre-configured with custom headers to capture
+    
+    with tracer.start_active_span('test'):
+        result = requests.get(testenv["fastapi_server"] + '/response_headers')
+
+    assert result.status_code == 200
```
Ferenc-, line 378: This TS was unfortunately also full with non-unit test assert, but perhaps we could use `self.assertEqual` and further below.

Ferenc- (changes requested):
Looking good. I have requested a couple of small changes.

instana/instrumentation/asgi.py
pvital: +1 for Ferenc's comment. You can follow the same as you did for Django's PR.

pvital (approved):
Besides the one comment, It looks good to me.

Ferenc- (approved):
Approved.

PR #489, opened 2023-12-28
Django: capture responseHeadersOnEntrySpans

GSVarsha:
Django: capture responseHeadersOnEntrySpans

### DIFF
--- a/instana/instrumentation/django/middleware.py
+++ b/instana/instrumentation/django/middleware.py
@@ -29,6 +29,19 @@ class InstanaMiddleware(MiddlewareMixin):
         super(InstanaMiddleware, self).__init__(get_response)
         self.get_response = get_response
 
+    def _extract_custom_headers(self, span, headers, format):
+        try:
+            if agent.options.extra_http_headers is not None:
+                for custom_header in agent.options.extra_http_headers:
+                    # Headers are available in this format: HTTP_X_CAPTURE_THIS
+                    django_header = ('HTTP_' + custom_header.upper()).replace('-', '_') if format else custom_header
+
+                    if django_header in headers:
+                        span.set_tag("http.header.%s" % custom_header, headers[django_header])
+
+        except Exception:
+            logger.debug("extract_custom_headers: ", exc_info=True)
+
     def process_request(self, request):
         try:
             env = request.environ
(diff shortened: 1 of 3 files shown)


instana/instrumentation/django/middleware.py
```diff
@@ -29,19 +29,27 @@ def __init__(self, get_response=None):
         super(InstanaMiddleware, self).__init__(get_response)
         self.get_response = get_response
 
+    def _extract_custom_headers(self, span, headers, format):
+        try:
+            if agent.options.extra_http_headers is not None:
```
Ferenc-, line 34: This check should be inverted for an early return and should go touside of the `try` block.

tests/apps/app_django.py
```diff
@@ -125,10 +125,17 @@ def complex(request):
     return HttpResponse('Stan wuz here!')
 
 
+def response_with_headers(request):
+    response = HttpResponse(content_type='')
+    response['X-Capture-This-Too'] = 'this too'
+    return response
```
Ferenc-, line 131:
```suggestion
    return HttpResponse('Stan wuz here with headers!', headers={'X-Capture-This-Too': 'this too'})
```

Ferenc-: I think it is still safer if we cover the capture of a list of headers, and not just a single header. Although the `test_request_header_capture` tests with a list and currently that covers the same codepath, but that is an implementational detail which at least theoretically could change in the future.

tests/frameworks/test_django.py
```diff
@@ -306,6 +307,49 @@ def test_custom_header_capture(self):
         assert "X-Capture-That" in django_span.data["http"]["header"]
         self.assertEqual("that", django_span.data["http"]["header"]["X-Capture-That"])
 
+        agent.options.extra_http_headers = original_extra_http_headers
+
+    def test_response_header_capture(self):
+        # Hack together a manual custom headers list
+        original_extra_http_headers = agent.options.extra_http_headers
+        agent.options.extra_http_headers = [u'X-Capture-This-Too']
```
Ferenc-, line 315: I think it is safer to test with more than one headers at a time.

tests/frameworks/test_django.py
```diff
@@ -306,6 +307,49 @@ def test_custom_header_capture(self):
...
+
+        with tracer.start_active_span('test'):
+            response = self.http.request('GET', self.live_server_url + '/response_with_headers')
+
+        assert response
```
Ferenc-, line 320: I know this was all over this TS, but this should really be `self.assertTrue`.

tests/frameworks/test_django.py
```diff
@@ -306,6 +307,49 @@ def test_custom_header_capture(self):
...
+
+        self.assertEqual(None, django_span.ec)
+        self.assertIsNone(django_span.stack)
+
+        self.assertEqual('/response_with_headers', django_span.data["http"]["url"])
+        self.assertEqual('GET', django_span.data["http"]["method"])
+        self.assertEqual(200, django_span.data["http"]["status"])
+        self.assertEqual('^response_with_headers$', django_span.data["http"]["path_tpl"])
+
+        assert "X-Capture-This-Too" in django_span.data["http"]["header"]
```
Ferenc-, line 348:
```suggestion
        self.assertIn( "X-Capture-This-Too", django_span.data["http"]["header"])
```

tests/frameworks/test_django.py
```diff
@@ -306,6 +307,49 @@ def test_custom_header_capture(self):
...
+        self.assertEqual("this too", django_span.data["http"]["header"]["X-Capture-This-Too"])
+
```
Ferenc-, line 350: If we had another header to capture, then we could validate it here.

Ferenc- (changes requested):
Good job! I have suggested a few minor changes.

Ferenc- (approved):
LGTM

pvital (approved):
It looks good to me.

PR #491, opened 2024-01-02
Flask: capture responseHeadersOnEntrySpans

GSVarsha:
Flask: capture responseHeadersOnEntrySpans

### DIFF
--- a/instana/instrumentation/flask/vanilla.py
+++ b/instana/instrumentation/flask/vanilla.py
@@ -13,23 +13,9 @@ import wrapt
 from ...log import logger
 from ...singletons import agent, tracer
 from ...util.secrets import strip_secrets_from_query
+from common import extract_custom_headers
 
 path_tpl_re = re.compile('<.*>')
-
-
-def extract_custom_headers(span, headers, format):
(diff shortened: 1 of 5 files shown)


instana/instrumentation/flask/vanilla.py
```diff
@@ -17,6 +17,21 @@
 path_tpl_re = re.compile('<.*>')
 
 
+def extract_custom_headers(span, headers, format):
```
Ferenc-, line 20: If the extract function is the same for both vanilla and with_blinker, then we could put this function into a separate file and include it here and into with_blinker.

GSVarsha: @Ferenc- Is it okay if I add the function in [this](https://github.com/instana/python-sensor/blob/master/instana/instrumentation/flask/common.py) file?

Ferenc-: I think it could fit there, yes.

Ferenc- (approved):
LGTM

pvital (approved):
It looks good to me.

PR #497, opened 2024-01-22
boto3: initial code for capturing request headers

GSVarsha:
Following [this](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html#before-call) boto3 documentation to capture the request headers.


GSVarsha:
Made changes in a different branch. 
PR [#499](https://github.com/instana/python-sensor/pull/499)

PR #498, opened 2024-01-29
fix(tests): Adapt to moto 5

Ferenc-:


tests/clients/boto3/test_boto3_lambda.py
```diff
@@ -7,7 +7,12 @@
 import boto3
 import pytest
 
-from moto import mock_lambda
+# TODO: Remove branching when we drop support for Python 3.7
+import sys
+if sys.version_info.minor >= 8:
```
GSVarsha, line 12: Same here and in other `test_boto3_*.py` files.

GSVarsha (approved):
Looks good to me!

PR #499, opened 2024-01-29
Enhance boto3 instrumentation to capture HTTP headers in all four flavours

GSVarsha:
## What

Enhance boto3 instrumentation to capture HTTP headers in all four flavours

### References

- [Kanbanize Card](https://instana.kanbanize.com/ctrl_board/197/cards/150067/details/)

## Why

According to [our spec](https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#capturing-headers-on-http-spans), tracers SHOULD capture all four flavours of HTTP headers.

### DIFF
--- a/instana/instrumentation/boto3_inst.py
+++ b/instana/instrumentation/boto3_inst.py
@@ -17,7 +17,7 @@ try:
     from boto3.s3 import inject
 
     def extract_custom_headers(span, headers):
-        if agent.options.extra_http_headers is None:
+        if agent.options.extra_http_headers is None or headers is None:
             return
         try:
             for custom_header in agent.options.extra_http_headers:
--- a/instana/instrumentation/boto3_inst.py
+++ b/instana/instrumentation/boto3_inst.py
@@ -46,16 +46,16 @@ try:
             logger.debug("non-fatal lambda_inject_context: ", exc_info=True)
 
 
-    @wrapt.patch_function_wrapper("botocore.hooks", "HierarchicalEmitter.emit")
-    def emit_request_created_with_instana(wrapped, instance, args, kwargs):
+    @wrapt.patch_function_wrapper("botocore.auth", "SigV4Auth.add_auth")
+    def emit_add_auth_with_instana(wrapped, instance, args, kwargs):
         active_tracer = get_active_tracer()
-        
-        # If we're not tracing or the event emitted is not request-created, just return;
-        if active_tracer is None or args[0].split(".")[0] != "request-created":
+
+        # If we're not tracing, just return;
+        if active_tracer is None:
             return wrapped(*args, **kwargs)
         
         span = active_tracer.active_span
-        extract_custom_headers(span, kwargs["request"].headers)
+        extract_custom_headers(span, args[0].headers)
 
         return wrapped(*args, **kwargs)
 
--- a/instana/instrumentation/boto3_inst.py
+++ b/instana/instrumentation/boto3_inst.py
@@ -46,19 +46,19 @@ try:
             logger.debug("non-fatal lambda_inject_context: ", exc_info=True)
 
 
-    @wrapt.patch_function_wrapper("botocore.hooks", "HierarchicalEmitter.emit_until_response")
-    def emit_until_response_with_instana(wrapped, instance, args, kwargs):
+    @wrapt.patch_function_wrapper("botocore.hooks", "HierarchicalEmitter.emit")
+    def emit_request_created_with_instana(wrapped, instance, args, kwargs):
         active_tracer = get_active_tracer()
         
-        # If we're not tracing or the event emitted is not before-call, just return;
-        if active_tracer is None or args[0].split(".")[0] != "before-call":
+        # If we're not tracing or the event emitted is not request-created, just return;
+        if active_tracer is None or args[0].split(".")[0] != "request-created":
             return wrapped(*args, **kwargs)
         
         span = active_tracer.active_span
-        if "custom_request_headers" in kwargs["context"]:
-            extract_custom_headers(span, kwargs["context"]["custom_request_headers"])
+        extract_custom_headers(span, kwargs["request"].headers)
 
         return wrapped(*args, **kwargs)
+
 
     @wrapt.patch_function_wrapper('botocore.client', 'BaseClient._make_api_call')
     def make_api_call_with_instana(wrapped, instance, arg_list, kwargs):
--- a/instana/instrumentation/boto3_inst.py
+++ b/instana/instrumentation/boto3_inst.py
@@ -8,13 +8,24 @@ import wrapt
 import inspect
 
 from ..log import logger
-from ..singletons import tracer
+from ..singletons import tracer, agent
 from ..util.traceutils import get_active_tracer
 
 try:
     import opentracing as ot
     import boto3
     from boto3.s3 import inject
+
+    def extract_custom_headers(span, headers):
+        if agent.options.extra_http_headers is None:
+            return
+        try:
+            for custom_header in agent.options.extra_http_headers:
+                if custom_header in headers:
+                    span.set_tag("http.header.%s" % custom_header, headers[custom_header])
+
+        except Exception:
+            logger.debug("extract_custom_headers: ", exc_info=True)
 
 
     def lambda_inject_context(payload, scope):
@@ -34,6 +45,20 @@ try:
         except Exception:
             logger.debug("non-fatal lambda_inject_context: ", exc_info=True)
 
+
+    @wrapt.patch_function_wrapper("botocore.hooks", "HierarchicalEmitter.emit_until_response")
+    def emit_until_response_with_instana(wrapped, instance, args, kwargs):
+        active_tracer = get_active_tracer()
+        
+        # If we're not tracing or the event emitted is not before-call, just return;
+        if active_tracer is None or args[0].split(".")[0] != "before-call":
+            return wrapped(*args, **kwargs)
+        
+        span = active_tracer.active_span
+        if "custom_request_headers" in kwargs["context"]:
+            extract_custom_headers(span, kwargs["context"]["custom_request_headers"])
+
+        return wrapped(*args, **kwargs)
 
     @wrapt.patch_function_wrapper('botocore.client', 'BaseClient._make_api_call')
     def make_api_call_with_instana(wrapped, instance, arg_list, kwargs):
@@ -76,6 +101,8 @@ try:
                         status = http_dict.get('HTTPStatusCode')
                         if status is not None:
                             scope.span.set_tag('http.status_code', status)
+                        headers = http_dict.get('HTTPHeaders')
+                        extract_custom_headers(scope.span, headers)
 
                 return result
             except Exception as exc:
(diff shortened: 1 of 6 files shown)


tests/clients/boto3/test_boto3_s3.py
```diff
@@ -2,277 +2,392 @@
...
-    filter = lambda span: span.n == "boto3"
-    boto_span = get_first_span_by_filter(spans, filter)
-    assert (boto_span)
+    def test_s3_list_buckets(self):
+        with tracer.start_active_span('test'):
+            self.s3.list_buckets()
 
-    assert (boto_span.t == test_span.t)
-    assert (boto_span.p == test_span.s)
+        result = self.s3.list_buckets()
```
Ferenc-, line 93: This shouldn't be needed here. This should go onto line 91.

tests/clients/boto3/test_boto3_s3.py
```diff
@@ -2,277 +2,392 @@
...
+        self.assertEqual(boto_span.data['http']['status'], 200)
+        self.assertEqual(boto_span.data['http']['method'], 'POST')
+        self.assertEqual(boto_span.data['http']['url'], 'https://s3.amazonaws.com:443/CreateBucket')
+
+        self.assertIn("X-Capture-This", boto_span.data["http"]["header"])
+        self.assertEqual("this", boto_span.data["http"]["header"]["X-Capture-This"])
+        self.assertIn("X-Capture-That", boto_span.data["http"]["header"])
+        self.assertEqual("that", boto_span.data["http"]["header"]["X-Capture-That"])
+            
+        agent.options.extra_http_headers = original_extra_http_headers
```
Ferenc-, line 333: Ideally this would go into the `tearDown`, because in case of any exception or assertion failure, this won't run and has the potential to impact other test cases.

tests/clients/boto3/test_boto3_s3.py
```diff
@@ -2,277 +2,392 @@
...
+
+        # Create a function that sets the custom headers in the after-call event.
+        def modify_after_call_args(parsed, **kwargs):
+            parsed['ResponseMetadata']['HTTPHeaders'].update(response_headers)
+
+        # Register the function to an event
+        event_system.register('after-call.s3.CreateBucket', modify_after_call_args)
+
+        with tracer.start_active_span('test'):
+            result = self.s3.create_bucket(Bucket="aws_bucket_name")
```
Ferenc-, line 357: This `result` is overwritten immediately, so we might not need this assignment here.

instana/instrumentation/boto3_inst.py
```diff
@@ -35,6 +46,20 @@ def lambda_inject_context(payload, scope):
...
+    @wrapt.patch_function_wrapper("botocore.hooks", "HierarchicalEmitter.emit_until_response")
+    def emit_until_response_with_instana(wrapped, instance, args, kwargs):
+        active_tracer = get_active_tracer()
+        
+        # If we're not tracing or the event emitted is not before-call, just return;
+        if active_tracer is None or args[0].split(".")[0] != "before-call":
+            return wrapped(*args, **kwargs)
+        
+        span = active_tracer.active_span
+        if "custom_request_headers" in kwargs["context"]:
```
Ferenc-, line 58: 1. As it turned out there is no official way for setting custom request headers on boto3.
See https://github.com/boto/boto3/issues/2889 & https://github.com/boto/botocore/issues/2050.

2. Any header, that could be used for an HTTP connection (including the built-in ones or any custom ones) are in the `kwargs['params']['headers']` dictionary in this code scope. That is what we could extract from. Anything else is non relevant from tracing perspective. But the default request headers are rather minimal, so minimal value for the customer.

3. Apparently the most accepted "workaround" for setting custom request headers is this one:
https://stackoverflow.com/questions/58828800/adding-custom-headers-to-all-boto3-requests
but there could be variations.

I am fine with either:
A, documenting the resons, and completely omitting the extraction from request headers.
B, only extract from default headers in  `kwargs['params']['headers']`

Ferenc- (changes requested):
The request header extraction should either focus on headers that paricipate in the HTTP call, or should be omitted.

tests/clients/boto3/test_boto3_s3.py
```diff
@@ -2,277 +2,392 @@
...
+        self.recorder.clear_spans()
+        self.aws_credentials()
+        self.mock = mock_aws()
+        self.mock.start()
+        self.s3 = boto3.client('s3', region_name='us-east-1')
+
+
+    def tearDown(self):
+        # Stop Moto after each test
+        self.mock.stop()
```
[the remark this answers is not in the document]
GSVarsha: Done!

instana/instrumentation/boto3_inst.py
```diff
@@ -35,6 +46,20 @@ def lambda_inject_context(payload, scope):
             logger.debug("non-fatal lambda_inject_context: ", exc_info=True)
 
 
+    @wrapt.patch_function_wrapper("botocore.hooks", "HierarchicalEmitter.emit")
```
Ferenc-, line 49: This hook gets called around a dozen of times per each request, which might have some performance impacts.
Regardless of the performance aspect, the main issue is that this simply doesn't always work.
Since there is no offical way of adding headers, the customer might get the idea [from here](https://stackoverflow.com/a/66097613), to add the headers with with `event_system.register_first('before-sign.s3.CreateBucket', _add_header)`, and in that case unfortunately this hook is too early and doesn't seem to work.

tests/clients/boto3/test_boto3_s3.py
```diff
@@ -296,8 +296,15 @@ def test_request_header_capture(self):
         def add_custom_header_before_call(params, **kwargs):
             params['headers'].update(request_headers)
 
-        # Register the function to an event.
-        event_system.register('before-call.s3.CreateBucket', add_custom_header_before_call)
+        # # Register the function to before-call event.
+        # event_system.register('before-call.s3.CreateBucket', add_custom_header_before_call)
```
Ferenc-, line 300: We could put this into a new TC.

Ferenc- (approved):
Very well done! Hats off.

instana/instrumentation/boto3_inst.py
GSVarsha: Made changes to extract headers from `add_auth` method of `botocore.auth`. This way, the hook gets called only once per api_call and adapts to both `before-call` and `before-sign` events.

tests/clients/boto3/test_boto3_s3.py
GSVarsha: Done!

tests/clients/boto3/test_boto3_lambda.py
```diff
@@ -2,74 +2,256 @@
...
 import boto3
-import pytest
-
 # TODO: Remove branching when we drop support for Python 3.7
-import sys
-if sys.version_info >= (3, 8):
+from sys import version_info
+if version_info >= (3, 8):
   from moto import mock_aws
 else:
```
Ferenc-, line 13: I think, this `else` branch should be superfluous, since the only TC in this file is skipped anyway.

instana/instrumentation/boto3_inst.py
```diff
@@ -8,14 +8,25 @@
...
+from ..singletons import tracer, agent
 from ..util.traceutils import get_active_tracer
 
 try:
     import opentracing as ot
     import boto3
     from boto3.s3 import inject
 
+    def extract_custom_headers(span, headers):
+        if agent.options.extra_http_headers is None:
```
Ferenc-, line 20: Just to be safe I would also add `or headers is None`. Unlikely to happen, but theoretically still possible.

tests/clients/boto3/test_boto3_lambda.py
```diff
@@ -2,74 +2,256 @@
...
+
+
+    def test_lambda_invoke(self):
+        with tracer.start_active_span('test'):
+            result = self.aws_lambda.invoke(FunctionName=self.function_name, Payload=json.dumps({"message": "success"}))
+
+        self.assertEqual(result["StatusCode"], 200)
+        result_payload = json.loads(result["Payload"].read().decode("utf-8"))
+        self.assertIn("message", result_payload)
+        self.assertEqual("success", result_payload["message"])
```
Ferenc-, line 43: Perhaps this could be simplified and improved like this?:
````
expected_payload = {"message": "success"}
with tracer.start_active_span('test'):
    result = self.aws_lambda.invoke(FunctionName=self.function_name, Payload=json.dumps(expected_payload))
self.assertEqual(result["StatusCode"], 200)
result_payload = json.loads(result["Payload"].read().decode("utf-8"))
self.assertDictEqual(expected_payload, result_payload)
````

File: tests/clients/boto3/test_boto3_lambda.py [hunk 11]
```diff
@@ -2,74 +2,256 @@
...
+        # Register the function to before-call event.
+        event_system.register('before-call.lambda.Invoke', add_custom_header_before_call)
+
+        with tracer.start_active_span('test'):
+            result = self.aws_lambda.invoke(FunctionName=self.function_name, Payload=json.dumps({"message": "success"}))
+
+        self.assertEqual(result["StatusCode"], 200)
+        result_payload = json.loads(result["Payload"].read().decode("utf-8"))
+        self.assertIn("message", result_payload)
+        self.assertEqual("success", result_payload["message"])
```
Ferenc-, line 98: Same as before, if possible, then let's make use of `assertDictEqual`.

tests/clients/boto3/test_boto3_lambda.py
```diff
@@ -2,74 +2,256 @@
...
+        # Register the function to before-sign event.
+        event_system.register_first('before-sign.lambda.Invoke', add_custom_header_before_sign)
+
+        with tracer.start_active_span('test'):
+            result = self.aws_lambda.invoke(FunctionName=self.function_name, Payload=json.dumps({"message": "success"}))
+
+        self.assertEqual(result["StatusCode"], 200)
+        result_payload = json.loads(result["Payload"].read().decode("utf-8"))
+        self.assertIn("message", result_payload)
+        self.assertEqual("success", result_payload["message"])
```
Ferenc-, line 161: Same as before, if possible, then let's make use of assertDictEqual.

tests/clients/boto3/test_boto3_lambda.py
GSVarsha: Yeah, I initially began with this approach, but it turns out the payload contains additional headers as well.

```python
{'message': 'success', 'X-INSTANA-L': '1', 'traceparent': '00-00000000000000006b2ac5343fa63053-ce05b7082d194345-01', 'tracestate': 'in=6b2ac5343fa63053;ce05b7082d194345', 'X-INSTANA-T': '6b2ac5343fa63053', 'X-INSTANA-S': 'ce05b7082d194345'}
```

tests/clients/boto3/test_boto3_lambda.py
GSVarsha: Addressed [here](https://github.com/instana/python-sensor/pull/499#discussion_r1481136233)

PR #500, opened 2024-02-07
fix(test_boto3) :Remove explicit credential mocking

GSVarsha:
- Remove explicit credential mocking
- Minor variable name fixes


GSVarsha:
> The explicit credential mocking is still the recommended action by the documentation: https://docs.getmoto.org/en/latest/docs/getting_started.html#recommended-usage But by now I think everyone runs these these tests in a containers anyway, and the TCs run with the mock decorator anyway, so I think it is still OK.

Thank you for letting me know!

PR #503, opened 2024-02-13
refactor: Remove superfluous absolute_import

Ferenc-:
This import was only relevant for Python 2.5 & 2.6. Beginning with 2.7 this was the default behaviour anyway, and now that the codebase only supports 3.7 and above, it is about time to get rid of this.


pvital (approved):
It looks good to me.  

I was thinking of doing the same. :-D Nice, we have it now.

PR #504, opened 2024-02-14
refactor(tests): Remove superfluous imports

GSVarsha:
Follow up PR for #503


instana/collector/base.py
```diff
@@ -13,10 +13,7 @@
 from ..util import every, DictionaryOfStan
 
 
-if sys.version_info.major == 2:
-    import Queue as queue
-else:
-    import queue # pylint: disable=import-error
+import queue # pylint: disable=import-error
```
Ferenc-, line 16: Is the comment needed?

instana/util/ids.py
```diff
@@ -11,13 +10,7 @@
...
 BAD_ID = "BADCAFFE"  # Bad Caffe
 
-PY2 = sys.version_info[0] == 2
-PY3 = sys.version_info[0] == 3
-
-if PY2:
-    string_types = basestring
-else:
-    string_types = str
+string_types = str
```
Ferenc-, line 13: I think we should just use str everywhere?

Ferenc- (changes requested):
Looking good. Perhaps a few extra lines could be shedded.

instana/collector/base.py
GSVarsha: Keeping the `pylint` comment as discussed.

Ferenc- (approved):
LGTM

pvital (approved):
It looks good to me.

PR #505, opened 2024-02-14
Sanic: capture responseHeadersOnEntrySpans

GSVarsha:
Sanic: capture responseHeadersOnEntrySpans

### DIFF
--- a/instana/instrumentation/sanic_inst.py
+++ b/instana/instrumentation/sanic_inst.py
@@ -44,6 +44,7 @@ try:
                 span.set_tag('http.status_code', status_code)
 
             if response.headers is not None:
+                extract_custom_headers(span, response.headers)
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -348,4 +348,4 @@ workflows:
             - python312
             - py39cassandra
             - py39couchbase
-            - py39gevent
+            - py39gevent
```
Ferenc-, line 351: Why the change here? In general we prefer newlines at the end of the file.

GSVarsha: I think I removed it by mistake. Fixed this in the latest commit.

pvital (approved):
It looks good to me.

Ferenc- (approved):
Approved

PR #506, opened 2024-02-14
Add .tekton

Ferenc-:


.tekton/github-set-status-task.yaml
```diff
@@ -0,0 +1,38 @@
...
+             -H "Accept: application/vnd.github+json" \
+             -H "X-GitHub-Api-Version: 2022-11-28" \
+             -H "Authorization: Bearer $(cat /etc/github-set-status/token)" \
+             "https://api.github.com/repos/instana/python-sensor/statuses/${SHA}" \
+             -d "{
+                  \"state\":\"${STATE}\",
+                  \"target_url\":\"http://localhost:8001/api/v1/namespaces/tekton-pipelines/services/tekton-dashboard:http/proxy/#/namespaces/default/pipelineruns/\",
+                  \"description\":\"Tekton build is in state: ${STATE}!\",
+                  \"context\":\"Tekton\"
+                 }"
```
GSVarsha, line 38: For better readability, can we add `-H "Content-Type: application/json"` header to explicitly specify the content type as JSON and avoid adding escape characters?

```suggestion
        curl -L -X POST \
               -H "Accept: application/vnd.github+json" \
               -H "X-GitHub-Api-Version: 2022-11-28" \
               -H "Authorization: Bearer $(cat /etc/github-set-status/token)" \
               -H "Content-Type: application/json" \
               -d '{
                    "state": "'${STATE}'",
                    "target_url": "http://localhost:8001/api/v1/namespaces/tekton-pipelines/services/tekton-dashboard:http/proxy/#/namespaces/default/pipelineruns/",
                    "description": "Tekton build is in state: '${STATE}'!",
                    "context": "Tekton"
                   }' \
              "https://api.github.com/repos/instana/python-sensor/statuses/${SHA}"

```

.tekton/task.yaml
```diff
@@ -0,0 +1,167 @@
...
+  sidecars:
+    - name: google-cloud-pubsub
+      image: egymgmbh/pubsub-emulator
+      command:
+        - /init.sh
+        - test-project
+        - test-topic
+        - test-subscription
+    - name: mariadb
+      image: mariadb:11.2.3
```
pvital, line 125: Why not remove the version (container tag) and use the most updated version here? With that, we can always guarantee to trace against the most updated server versions. 

```suggestion
      image: mariadb
```

.tekton/task.yaml
```diff
@@ -0,0 +1,167 @@
...
+      env:
+        - name: MYSQL_ROOT_PASSWORD # or MARIADB_ROOT_PASSWORD
+          value: passw0rd
+        - name: MYSQL_DATABASE # or MARIADB_DATABASE
+          value: instana_test_db
+    - name: mongo
+      image: mongo:4.2.3
```
pvital, line 132: As mentioned before, tracing against the most updated server versions would be good.

```suggestion
      image: mongo
```

.tekton/task.yaml
```diff
@@ -0,0 +1,167 @@
...
+    - name: postgres
+      image: postgres:9.6.24
```
pvital, line 134: As mentioned before, tracing against the most updated server versions would be good.

```suggestion
      image: postgres
```

.tekton/task.yaml
```diff
@@ -0,0 +1,167 @@
...
+          value: instana_test_db
+      readinessProbe:
+        exec:
+          command:
+          - sh
+          - -c
+          - pg_isready --host 127.0.0.1 --port 5432 --dbname=${POSTGRES_DB}
+        timeoutSeconds: 10
+    - name: redis
+      image: redis:7.2.4
```
pvital, line 150: As mentioned before, tracing against the most updated server versions would be good. 

```suggestion
      image: redis
```

.tekton/task.yaml
```diff
@@ -0,0 +1,167 @@
...
+    - name: rabbitmq
+      image: rabbitmq:3.12.12
```
pvital, line 152: As mentioned before, tracing against the most updated server versions would be good.

```suggestion
      image: rabbitmq
```

pvital (changes requested):

.tekton/github-set-status-task.yaml
Ferenc-: Valid point. I have also sorted the headers.

.tekton/task.yaml
Ferenc-: For example the new pulls now have already exhausted some PQS limits when pulling from dockerhub:
`The sidecar "mongo" in TaskRun "python-tracer-pr-tekton-test-c2d58b8-unittest-default-1" failed to pull the image "". The pod errored with the message: "Back-off pulling image "mongo"."`
Among the pod events one can see:
````
  Normal  Killing    66s    kubelet            Container sidecar-mongo definition changed, will be restarted
  Normal  Killing    66s    kubelet            Container sidecar-postgres definition changed, will be restarted
````

Comment (Ferenc-) [reply .tekton/task.yaml]: I will experiment with a pull through proxy for dockerhub, that should help.

Comment (pvital) [reply .tekton/task.yaml]: I understand Tekton's recommendation, and, IMHO, I agree with that in parts. 

I guess we can then use the digest of the latest images that work for us, keep tracking for the changes, and create PRs to update the digests if necessary.

File: .tekton/github-webhook-ingress.yaml [hunk 7]
```diff
@@ -0,0 +1,20 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: github-webhook-ingress
+spec:
+  ingressClassName: public-iks-k8s-nginx
+  tls:
+  - hosts:
+    - <ENTER_YOUR_DOMAIN_NAME_HERE>
```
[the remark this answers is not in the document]
Ferenc-: I am only aware of GitHub secret variables in the context of the GitHub Actions CI, and this code under `.tekton` is not part of any GitHub Action code, so I would say it doesn't fit.

.tekton/task.yaml
Ferenc-: All right then I am changing to digest here.

pvital (approved):
It looks good to me.

PR #507, opened 2024-02-15
Enable Asyncio support.

pvital:
As we support now only the Python runtime >= 3.7, the usage of the ContextVarsScopeManager as Tracer's scope manager is indicated to provide automatic Span propagation from parent coroutines, tasks and scheduled in event loop callbacks to their children.


instana/instrumentation/asyncio.py
```diff
@@ -14,27 +14,33 @@
     @wrapt.patch_function_wrapper('asyncio','ensure_future')
     def ensure_future_with_instana(wrapped, instance, argv, kwargs):
         if config['asyncio_task_context_propagation']['enabled'] is False:
-            return wrapped(*argv, **kwargs)
+            from opentracing.scope_managers.contextvars import no_parent_scope
+            with no_parent_scope():
```
Ferenc-, line 18: Why the no_parent_scope? I think we would be very much interested in NOT breaking the span propagation.

instana/singletons.py
```diff
@@ -94,11 +94,10 @@ def set_agent(new_agent):
 # this package.
 tracer = InstanaTracer(recorder=span_recorder)
 
-if sys.version_info >= (3, 4):
+if sys.version_info >= (3, 7):
```
Ferenc-, line 97: Not necessarily the topic of this PR, but this whole if statement should be removed.

Ferenc- (changes requested):
In general I see the same tests passing before and after this change.
In order to see that this change has an added value, perhaps we could see some more tests passing? We could enable tornado 6.4 tests and see if this really enables something new which was not possible before.

instana/singletons.py
pvital: True. I will change it.

instana/instrumentation/asyncio.py
pvital: Yes, but this block runs only when I ask to break context propagation. 
I am using it to guarantee it will not be propagated; otherwise, this `if` block (and the whole function, IMHO) is useless.

pvital:
> In general I see the same tests passing before and after this change. In order to see that this change has an added value, perhaps we could see some more tests passing? We could enable tornado 6.4 tests and see if this really enables something new which was not possible before.

@Ferenc-, IMO, for a `refactor` commit, after changing something that creates a difference in the future execution, having all UT passing is a good signal that the code's functionality remains the same even after the refactoring changes.

Regarding the support for Tornado > 6.0, this will demand more changes, and I am working on it in a different branch at the moment and will open a new PR for that later.

instana/instrumentation/asyncio.py
Ferenc-: You are right, I missed the point. This is that underdocumented one of a kind mummy that the customer is supposed to use manually with `from instana.configurator import config`.  I wonder if there is even a single customer making use of this.

Ferenc- (approved):
I do see a lot of value in the findings proven by this PR. Perhaps I was to worried at first.

instana/instrumentation/asyncio.py
```diff
@@ -14,27 +14,33 @@
     @wrapt.patch_function_wrapper('asyncio','ensure_future')
     def ensure_future_with_instana(wrapped, instance, argv, kwargs):
         if config['asyncio_task_context_propagation']['enabled'] is False:
-            return wrapped(*argv, **kwargs)
+            from opentracing.scope_managers.contextvars import no_parent_scope
```
Ferenc-, line 17: I think this import could be moved out of this if statement:

1. Becase by default this branch is taken anyway, 
    most users have no idea that this branch can even be avoided, and this import is executed repeatedly, [which has an import statement overhead](https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Import_Statement_Overhead).
2.  The point of the import guard around this file is to find out import issues sooner, rather than later.
3.  The [Python style also prefers putting imports to the top](https://peps.python.org/pep-0008/#imports).
4.  It is less lines to read when trying to focus on the logic, and not on where certain names are coming from.

instana/instrumentation/asyncio.py
```diff
@@ -14,27 +14,33 @@
...
+            setattr(task, ACTIVE_ATTR, scope)
 
         return task
 
     if hasattr(asyncio, "create_task"):
         @wrapt.patch_function_wrapper('asyncio','create_task')
         def create_task_with_instana(wrapped, instance, argv, kwargs):
             if config['asyncio_task_context_propagation']['enabled'] is False:
-                return wrapped(*argv, **kwargs)
+                from opentracing.scope_managers.contextvars import no_parent_scope
```
Ferenc-, line 34: Same comments as above.

instana/instrumentation/asyncio.py
```diff
@@ -14,27 +14,33 @@
...
+                with no_parent_scope():
+                    return wrapped(*argv, **kwargs)
 
             scope = async_tracer.scope_manager.active
             task = wrapped(*argv, **kwargs)
 
             if scope is not None:
-                async_tracer.scope_manager._set_task_scope(scope, task=task)
+                from opentracing.scope_managers.constants import ACTIVE_ATTR
```
Ferenc-, line 42: Here as well.

Ferenc- (approved):

GSVarsha (approved):
My only suggestion would be to move the imports to the top of the file as mentioned in previous comments. Other than that, the PR looks good to me!

instana/instrumentation/asyncio.py
pvital: Got it. Moved.

instana/instrumentation/asyncio.py
pvital: Done.

instana/instrumentation/asyncio.py
pvital: Done.

Ferenc- (approved):
Looks good.

PR #508, opened 2024-02-15
Refectoring and improvements

pvital:
This PR contains two commits to refector and improve our code:

1. The first guarantee container images always come from the Docker registry. That prevents that podman, for example, running in RHEL systems, from downloading images from other registries.
2. To reduce the maintenance of the requirements used in our tests, this fix refers to other requirement files that have the same list of packages.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.8'
 services:
   redis:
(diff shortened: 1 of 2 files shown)


Ferenc- (approved):
LGTM

PR #509, opened 2024-02-18
Tornado Server: capture responseHeadersOnEntrySpans

GSVarsha:
tornado_server: capture responseHeadersOnEntrySpans

### DIFF
--- a/instana/instrumentation/tornado/server.py
+++ b/instana/instrumentation/tornado/server.py
@@ -23,6 +23,18 @@ try:
 
     setup_tornado_tracer()
 
+    def extract_custom_headers(span, headers):
+        if not agent.options.extra_http_headers or not headers:
+            return
+        try:
+            for custom_header in agent.options.extra_http_headers:
+                if custom_header in headers:
+                    span.set_tag("http.header.%s" % custom_header, headers[custom_header])
+
+        except Exception:
+            logger.debug("extract_custom_headers: ", exc_info=True)
+
+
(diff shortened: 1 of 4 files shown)


instana/instrumentation/tornado/server.py
```diff
@@ -23,6 +23,18 @@
 
     setup_tornado_tracer()
 
+    def extract_custom_headers(span, headers):
+        if agent.options.extra_http_headers is None:
```
Ferenc-, line 27: Theoretically `headers` could also be None not just the `extra_http_headers`, plus strictly speaking the early return would make sense in a broader set of cases. Not just in case of None, but in case of anything Falsy, like empty collections etc. So the `not` operator could cover more.

instana/instrumentation/tornado/server.py
```diff
@@ -80,15 +88,17 @@ def on_finish_with_instana(wrapped, instance, argv, kwargs):
...
+                extract_custom_headers(scope.span, instance._headers)
+
+                status_code = instance.get_status()
 
-            # Mark 500 responses as errored
-            if 500 <= status_code:
-                scope.span.mark_as_errored()
+                # Mark 500 responses as errored
+                if 500 <= status_code:
+                    scope.span.mark_as_errored()
```
Ferenc-, line 99: This extra inden

Ferenc- (approved):
Looks good.

pvital (approved):
It looks good to me.

Ferenc- (approved):
Approved.

PR #514, opened 2024-03-04
[FastAPI] Add TC for non-async methods

GSVarsha:
Add a non-async path operation function that makes a complex non-asynchronous request to check if all the outgoing calls made by the request are properly traced.

### DIFF
--- a/tests/frameworks/test_fastapi.py
+++ b/tests/frameworks/test_fastapi.py
@@ -7,7 +7,7 @@ import multiprocessing
 
 import requests
 
-from instana.singletons import tracer
+from instana.singletons import async_tracer
 from tests.apps.fastapi_app import launch_fastapi
 from ..helpers import testenv
 from ..helpers import get_first_span_by_filter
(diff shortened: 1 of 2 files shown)


pvital (approved):
It looks good to me.

tests/frameworks/test_fastapi.py
```diff
@@ -503,3 +503,54 @@ def test_response_header_capture(self):
         self.assertEqual("this too", asgi_span.data["http"]["header"]["X-Capture-This-Too"])
         self.assertIn("X-Capture-That-Too", asgi_span.data["http"]["header"])
         self.assertEqual("that too", asgi_span.data["http"]["header"]["X-Capture-That-Too"])
+
+    def test_non_async_function(self):
+        with tracer.start_active_span("test"):
```
Ferenc-, line 508: While the usage of the normal `tracer` instead of `async_tracer` is not new to this file,
since the `fastapi` instrumentation uses the `async_tracer`, I feel we should use `async_tracer` in the tests as well.

Ferenc- (changes requested):
Looks good, only the usage of `async_tracer` instead of `tracer` could be perhaps improved.

Ferenc- (approved):
Looks good.

GSVarsha:
New tests seem to be flaky on python 3.12 - investigating on the same

GSVarsha:
> ⚠️ New tests seem to be flaky on python 3.12 - investigating on the same

Span order was the issue. Used span filters to resolve the same.

Ferenc- (commented):
Looks good. Now that we are here, could we also add some TCs, that make explicit use of `from fastapi.concurrency import run_in_threadpool`?
I think that might be the area where we could pontentially find limitations pertaining to the tracer's thread boundary, and where we might have to document something.

pvital (changes requested):
It looks good to me, but your commit message is a bit messy. 

Could you `rebase` and `reword` your commit message before merging it, please?

pvital (approved):
It looks good to me.

PR #516, opened 2024-03-14
Initial 3.x

Ferenc-:


setup.py
```diff
@@ -45,6 +45,10 @@
                         'basictracer>=3.1.0',
                         'fysom>=2.1.2',
                         'opentracing>=2.3.0',
+                        'opentelemetry-api',
+                        'opentelemetry-distro',
+                        'opentelemetry-opentracing-shim',
+                        'opentelemetry-sdk',
```
pvital, line 51: These dependencies are not necessary. As described in the README file of the [opentelemetry-python](https://github.com/open-telemetry/opentelemetry-python?tab=readme-ov-file#install) project,

>  Libraries that produce telemetry data should only depend on `opentelemetry-api`.

Also, I suggest already removing the `opentracing` and `basictracer` packages - yes, the tests and everything else will break, but those errors can help us detect where changes will be necessary.

pvital (changes requested):

pvital:
I suggest to already change the version number to `3.0.0` or `3.0.0.dev`.

PR #522, opened 2024-04-26
feat: Add opt-in exit spans (part 1)

Ferenc-:


instana/util/traceutils.py
```diff
@@ -30,3 +30,16 @@ def get_active_tracer():
         # Do not try to log this with instana, as there is no active tracer and there will be an infinite loop at least
         # for PY2
         return None
+
+
+def get_tracer_tuple():
+    active_tracer = get_active_tracer()
+    if get_active_tracer():
```
GSVarsha, line 37: Since we have already defined `active_tracer = get_active_tracer()`
```suggestion
    if active_tracer:
```

instana/util/traceutils.py
```diff
@@ -30,3 +30,16 @@ def get_active_tracer():
...
+        return active_tracer, active_tracer.active_span, active_tracer.active_span.operation_name
+    elif agent.options.allow_exit_as_root:
+        return tracer, None, None
```
GSVarsha, line 40: Can we explicitly return a tuple in this function for better readability?
```suggestion
        return (tracer, None, None)
```

instana/util/traceutils.py
Ferenc-: Done

instana/util/traceutils.py
Ferenc-: Done

GSVarsha (approved):
Looks perfect to me!

PR #525, opened 2024-05-03
Build migration to pyproject.toml

pvital:
This PR suggests changing the current build and package files to the latest standard Python building system. It prevents issues with the Instana package not being installed or built if the package `wheel` is not installed in the running environment (see https://github.com/pypa/pip/issues/8559), which is common in fresh environments with recent `pip` versions. 

To solve that, the suggestion is to use `pyproject.toml` instead of `setup.py` as building and packaging files. [PEP 621](https://peps.python.org/pep-0621/) and [PEP 631](https://peps.python.org/pep-0631/) standardized `pyproject.toml` as the new normal for Python packages instead of `setup.py`. [PEP 517](https://peps.python.org/pep-0517/) and [PEP 660](https://peps.python.org/pep-0660/) created standards for Python build systems.

This PR:
1. moves all the Instana package files under the `src` directory;
2. removes the `setup.py`; and 
3. creates the `pyproject. toml` file to be the new build and package file. 
 
The new `pyproject. toml` file uses [Hatchling](https://packaging.python.org/en/latest/key_projects/#hatch) as the default build-backend, and it must be installed in your system to build the package.

```bash
brew install hatch    # or from https://hatch.pypa.io/latest/install/#build-system-availability
hatch --version        # to verify the installation
hatch build 
```


pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+keywords = [
+    "performance",
+    "opentracing",
+    "metrics",
+    "monitoring",
+    "tracing",
+    "distributed-tracing",
+]
+authors = [
+    { name = "Instana Team Python Engineers", email = "info@instana.com" },
```
Ferenc-, line 23: If `info@instana.com` still works, then its days should be numbered. We shouldn't use it anymore.

pvital: Got it. What do you think about the authors list @Ferenc-?

pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+license = "MIT"
+keywords = [
+    "performance",
+    "opentracing",
+    "metrics",
+    "monitoring",
+    "tracing",
+    "distributed-tracing",
+]
+authors = [
```
GSVarsha, line 22: Shouldn't the authors list contain the list of contributors of this repository or the current team members of python tracers?

If we decide to go generic, I would suggest the name to be something like `"Instana Team Python Tracer Engineers"`

Ferenc-: We could just reference https://github.com/orgs/instana/teams/eng-python

pyproject.toml
pvital: As the [specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#authors-maintainers) states that only the name is a valid entry, I removed the email address.

pyproject.toml
pvital: The [specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#authors-maintainers) states that:
> The name value MUST be a valid email name (i.e. whatever can be put as a name, before an email, in [RFC 822](https://datatracker.ietf.org/doc/html/rfc822.html)) and not contain commas.

So, as `Instana Team Python Tracer Engineers` is a valid email name, I modified it to this name.

pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+    "Intended Audience :: Information Technology",
+    "Intended Audience :: Science/Research",
+    "Intended Audience :: System Administrators",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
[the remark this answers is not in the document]
pvital: I only copied the identical versions stated in the `setup.py` file, but I can add them without problems.

Ferenc- (approved):
LGTM

PR #539, opened 2024-06-12
fix: Prevent crash with 'gevent' has no attribute 'version_info'

Ferenc-:
If the `gevent` module is not installed,
but there is a directory called `gevent`
or a `gevent.py` file in the current directory,
then that is already importable and the statment
`if sys.modules['gevent'].version_info < (1, 4):`
fails with
`AttributeError: module 'gevent' has no attribute 'version_info'`

This can happen easily since many libraries like `celery`, `grpc`, `opentracing` etc, deliver their own file called `gevent.py`.


GSVarsha (approved):
Looks good to me!

PR #542, opened 2024-06-12
chore: Drop 3.7 support

Ferenc-:
This is needed for multiple reasons,
most pressing is that the `importlib.metadata` is only available in 3.8, and that would be needed to fix deprecation warnings about `pkg_resources`.


GSVarsha (changes requested):
Should we consider removing 3.7 from [here](https://github.com/instana/python-sensor/blob/drop-37/bin/aws-lambda/build_and_publish_lambda_layer.py#L134) as well?

Ferenc-:
> Should we consider removing 3.7 from [here](https://github.com/instana/python-sensor/blob/drop-37/bin/aws-lambda/build_and_publish_lambda_layer.py#L134) as well?

Yes, that should have been removed [long ago](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html).

GSVarsha (changes requested):
Can we remove [this](https://github.com/instana/python-sensor/blob/drop-37/.tekton/run_unittests.sh#L22) because someone 3.7 is getting picked up by the [CI runs](http://localhost:8001/api/v1/namespaces/tekton-pipelines/services/tekton-dashboard:http/proxy/#/namespaces/default/pipelineruns/python-tracer-pr-drop-37-84e9256?pipelineTask=unittest-default&taskRunName=python-tracer-pr-drop-37-84e9256-unittest-default-0&step=unittest).

GSVarsha (changes requested):
Can we remove all the branching for 3.7?

Screenshot 2024-06-12 at 4 02 56 PM
Screenshot 2024-06-12 at 4 04 48 PM

pvital (changes requested):
Perhaps you can also remove the `tests/requirements-307.txt` file in this PR.

tests/requirements-308.txt
pvital, line 1: Isn't the `requirements.txt` file the one to use on top of Python 3.8 and 3.9? 
It would be better to delete the `requirements-307.txt` and keep the `requirements.txt` one. Don't you think?

pvital (changes requested):

tests/requirements-308.txt
Ferenc-: This wasn't needed, this remained in the commit by mistake. Removed now.

pvital (approved):
It looks good to me.

GSVarsha (approved):
LGTM

PR #545, opened 2024-06-28
fix: Set AutoTrace method when the webhook is active

Ferenc-:
So do not look for the sideffects of the removed injector.


pvital (approved):
It looks good to me.

PR #546, opened 2024-06-28
Autotrace gevent

Ferenc-:


Ferenc-:
Depends on https://github.com/instana/python-sensor/pull/545

pvital (approved):
It looks good for me.

Only remember to rebase first, since part of this solution is already solved by #545 .

PR #548, opened 2024-07-02
report(currency): update support policy

GSVarsha:
Update support policy in python currency report.


pvital (approved):
It looks good to me.

PR #554, opened 2024-07-14
report(currency): update tracer support policy 30->45

GSVarsha:
We have been asked to update our markdown reports to change the Support Policy column.
For anything marked 30days, we are asked to revise to 45days.


GSVarsha:
Made changes in https://github.com/instana/python-sensor/pull/555 .

PR #556, opened 2024-07-15
fix: Report data to agent

GSVarsha:
Fix the below error that hinders reporting tracer data to the agent.
```
2024-07-15 12:41:31,430: 51725 DEBUG instana: to_json non-fatal encoding issue: 
  File "/Users/varsha/Documents/GitHub/instana-repos/tracer-repos/python-sensor/src/instana/util/__init__.py", line 36, in to_json
    return json.dumps(obj, default=extractor, sort_keys=False, separators=(',', ':')).encode()
  File "/Users/varsha/.pyenv/versions/3.9.7/lib/python3.9/json/__init__.py", line 234, in dumps
    return cls(
  File "/Users/varsha/.pyenv/versions/3.9.7/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/varsha/.pyenv/versions/3.9.7/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
```
`span.k` was of type ENUM class attribute. Change it to Literal[Spankind.\<attribute\>] which is basically the integer value of the ENUM class attribute before sending the spans to the agent.

### DIFF
--- a/src/instana/collector/utils.py
+++ b/src/instana/collector/utils.py
@@ -3,16 +3,17 @@
 from typing import TYPE_CHECKING, List
 
 from opentelemetry.trace.span import format_span_id
+from opentelemetry.trace import SpanKind
 
 if TYPE_CHECKING:
     from instana.span.span import InstanaSpan
 
 
-def format_trace_and_span_ids(
+def format_span(
     queued_spans: List["InstanaSpan"],
 ) -> List["InstanaSpan"]:
     """
-    Format the Trace, Parent Span, and Span IDs of Spans to be a 64-bit
+    Format Span Kind and the Trace, Parent Span and Span IDs of the Spans to be a 64-bit
     Hexadecimal String instead of Integer before being pushed to a
     Collector (or Instana Agent).
     """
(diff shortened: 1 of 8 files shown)


pvital (changes requested):
As discussed in our call, I think we lose all the features and characteristics of an `Enun` class if we go to this change. 

Instead, let's try and check if we can use the `value` of the properties when creating the JSON payload to send to the Agent. With that, we can continue to use the good of an `Enun` class and follow the Instana's specification.

src/instana/collector/utils.py
```diff
@@ -3,16 +3,17 @@
...
+from opentelemetry.trace import SpanKind
 
 if TYPE_CHECKING:
     from instana.span.span import InstanaSpan
 
 
-def format_trace_and_span_ids(
+def format_span(
     queued_spans: List["InstanaSpan"],
 ) -> List["InstanaSpan"]:
```
pvital, line 14: This is not correct anymore. As what we have in the `collector.queued_spans` are `RegisteredSpan` or `SDKSpan` objects (see [here](https://github.com/instana/python-sensor/blob/a1329947a336047e4e04de378f81d625ada3eaa3/src/instana/recorder.py#L88)), we can change this to really reflect that. 

So, we can do like this:
```suggestion
def format_span(
    queued_spans: List[Union[RegisteredSpan, SDKSpan]],
) -> List[Union[RegisteredSpan, SDKSpan]]:
```
or this:
```suggestion
def format_span(
    queued_spans: List[Type["BaseSpan"]],
) -> List[Type["BaseSpan"]]:
```

pvital (changes requested):

pvital (approved):
It looks good to me

PR #558, opened 2024-07-21
OTel small fixes.

pvital:
- style: Update type hints.
  - Use Type["BaseAgent"] instead of a long Union list with all its inherited classes.
  - Update the return type hint for the Span duration property.
- fix(tests): Add Span duration unit tests.
  - Add new tests to cover the changes on the (readable)span duration property.
  - Fix the Span duration update when ending a Span.


GSVarsha (approved):
Looks good to me!

PR #560, opened 2024-07-23
Added unittests for Readable Span

CagriYonca:
Added unittests for ReadableSpan.


tests/test_span_readable.py
```diff
@@ -0,0 +1,65 @@
...
+) -> None:
+    span_name = "test-span"
+    timestamp = time.time_ns()
+    span = ReadableSpan(span_name, span_context)
+
+    assert span is not None
+    assert isinstance(span, ReadableSpan)
+    assert span.name == span_name
+
+    context = span.context
```
GSVarsha, line 22:
```suggestion
    span_context = span.get_span_context()
```

Just a small suggestion. Since `Context` object is different from a `SpanContext` object, I would suggest renaming the variable and explicitly calling the `span.get_span_context()` method instead of directly accessing the `span.context` attribute. This is just for clarity, the result will be the same as earlier.

tests/test_span_readable.py
```diff
@@ -0,0 +1,65 @@
...
+        span_context: SpanContext,
+        trace_id: int,
+        span_id: int,
+) -> None:
+    span_name = "test-span"
+    tp = Traceparent()
+    traceparent_id = "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"
+    _, _, parent_id, _ = tp.get_traceparent_fields(traceparent_id)
+    start_time = time.time_ns()
+    end_time = time.time_ns()
```
GSVarsha, line 51: You could also check if the duration is the difference between end time and start time.

tests/test_span_readable.py
pvital: +1 on Varsha's suggestion.

tests/test_span_readable.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_readablespan_with_params(
+        span_context: SpanContext,
+        trace_id: int,
+        span_id: int,
+) -> None:
+    span_name = "test-span"
+    tp = Traceparent()
+    traceparent_id = "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"
+    _, _, parent_id, _ = tp.get_traceparent_fields(traceparent_id)
```
pvital, line 49: It's not necessary to use `Traceparent()` here. Just set a hardcoded `parent_id` value and use it. It will help you understand the test better, and you can use it in the `assert` part.

tests/test_span_readable.py
```diff
@@ -0,0 +1,65 @@
...
+
+    assert span.start_time
+    assert isinstance(span.start_time, int)
+    assert span.start_time > timestamp
+    assert not span.end_time
+    assert not span.attributes
+    assert not span.events
+    assert not span.parent_id
+    assert span.duration == 0
+    assert span.status._status_code == StatusCode.UNSET
```
pvital, line 35: `Status` objects have a property called `status_code`. 
```suggestion
    assert span.status.status_code == StatusCode.UNSET
```

tests/test_span_readable.py
```diff
@@ -0,0 +1,65 @@
...
+        span_id: int,
+) -> None:
+    span_name = "test-span"
+    tp = Traceparent()
+    traceparent_id = "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"
+    _, _, parent_id, _ = tp.get_traceparent_fields(traceparent_id)
+    start_time = time.time_ns()
+    end_time = time.time_ns()
+    attributes = {'key': 'value'}
+    event_name = 'event'
```
pvital, line 53: Please, use double quotes (") for all String definitions.
```suggestion
    attributes = {"key": "value"}
    event_name = "event"
```

pvital (changes requested):
What is the coverage percentile of `readable_span.py` with your changes?

tests/test_span_readable.py
CagriYonca: duration variable never changes inside Readable Span, is it normal?

tests/test_span_readable.py
CagriYonca: Done.

tests/test_span_readable.py
CagriYonca: I figured out that I was working on different branch version. Now I pulled the last version of otel_migration, it's okay now

tests/test_span_readable.py
CagriYonca: ReadableSpan doesn't have get_span_context function

CagriYonca:
> What is the coverage percentile of `readable_span.py` with your changes?

Coverage for this file is 98%.

tests/test_span_readable.py
pvital: @CagriYonca, so the `get_span_context` method must be moved from `span.py` to the `readable_span.py` file.
Can you make that change as well (in a specific commit)?

CagriYonca: Done!

CagriYonca:
sample_span._events was not cleared after running test_populate_exit_span_data_log() case. I've manually cleared before moving to the next case.

PR #562, opened 2024-07-24
fix: prevent tracing internal requests.

pvital:
The urllib3 library, which we instrument, is used by the requests package, and this last one is used by Instana Python Tracer to connect with an Instana Agent.

When the INSTANA_ALLOW_EXIT_AS_ROOT environment variable is set, the internal requests are traced as EXIT_ROOT spans.

This commit prevents the internal requests from being traced by our lib.

### DIFF
--- a/src/instana/agent/host.py
+++ b/src/instana/agent/host.py
@@ -40,6 +40,10 @@ class HostAgent(BaseAgent):
     """
     AGENT_DISCOVERY_PATH = "com.instana.plugin.python.discovery"
     AGENT_DATA_PATH = "com.instana.plugin.python.%d"
+    AGENT_REQUEST_HEADERS = {
+        "Content-Type": "application/json",
+        "X-INSTANA-L": "0",
+    }
 
     def __init__(self):
         super(HostAgent, self).__init__()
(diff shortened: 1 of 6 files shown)


GSVarsha (approved):
Looks good to me!

CagriYonca (approved):

CagriYonca (approved):

src/instana/agent/host.py
```diff
@@ -176,7 +176,10 @@ def announce(self, discovery):
             url = self.__discovery_url()
             response = self.client.put(url,
                                        data=to_json(discovery),
-                                       headers={"Content-Type": "application/json"},
+                                       headers={
```
Ferenc-, line 179: These exact four lines are repeated 4 ish times, 5 if we count paritals. Perhaps it makes sense to use a variable?

Ferenc- (commented):
In general it looks pretty good to me. One minor comment about code duplication.
Plus it might make sense to write some UT in for example `test_host.py` which could ensure that announce and other requests have suppressions set for future changes.

tests/platforms/test_host.py
```diff
@@ -234,6 +234,38 @@ def test_announce_fails_with_missing_uuid(self, mock_requests_session_put):
...
+            name=test_process_name, 
+            args=test_process_args
+        )
+        payload = self.agent.announce(d)
+
+        self.assertIn('pid', payload)
+        self.assertEqual(test_pid, payload['pid'])
+
+        spans = self.span_recorder.queued_spans()
+        self.assertEqual(len(spans), 0)
```
Ferenc-, line 267: This testcase doesn't even enable `allow_exit_as_root`, and this is passing even without the fixes in `eb5f7bc`, because without `allow_exit_as_root`, there is no recording of parentless exit spans.

Ferenc- (changes requested):

PR #566, opened 2024-07-29
Added constraint for setuptools Python module

CagriYonca:
There is a common issue with the last version of setuptools. It looks like they've removed the test attribute from setuptools.command module. I found a solution from the Github issue below and implemented it to our pipeline.

It's because of a package named spyne: https://github.com/arskom/spyne
It's not being updated for a long time (7 months), has some conflicts with setuptools

https://github.com/pypa/setuptools/issues/4519


pvital (approved):
It looks good to me.

PR #567, opened 2024-07-29
fix: log message >WARN not showing up in the UI

GSVarsha:
#### Issue:
Log message >WARN not showing up in the UI

Screenshot 2024-07-29 at 8 18 24 PM

#### Fix:
Revert `span.data["event"]` to `span.data["log"]`

Screenshot 2024-07-29 at 4 30 33 PM

### DIFF
--- a/src/instana/span/registered_span.py
+++ b/src/instana/span/registered_span.py
@@ -139,8 +139,8 @@ class RegisteredSpan(BaseSpan):
         if span.name == "render":
             self.data["render"]["name"] = span.attributes.pop("name", None)
             self.data["render"]["type"] = span.attributes.pop("type", None)
-            self.data["event"]["message"] = span.attributes.pop("message", None)
-            self.data["event"]["parameters"] = span.attributes.pop("parameters", None)
+            self.data["log"]["message"] = span.attributes.pop("message", None)
+            self.data["log"]["parameters"] = span.attributes.pop("parameters", None)
         else:
             logger.debug("SpanRecorder: Unknown local span: %s" % span.name)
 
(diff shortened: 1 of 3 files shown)


src/instana/span/registered_span.py
```diff
@@ -139,8 +139,8 @@ def _populate_local_span_data(self, span) -> None:
         if span.name == "render":
             self.data["render"]["name"] = span.attributes.pop("name", None)
             self.data["render"]["type"] = span.attributes.pop("type", None)
-            self.data["event"]["message"] = span.attributes.pop("message", None)
-            self.data["event"]["parameters"] = span.attributes.pop("parameters", None)
+            self.data["log"]["message"] = span.attributes.pop("message", None)
+            self.data["log"]["parameters"] = span.attributes.pop("parameters", None)
```
pvital, line 143: Why are you assuming it's not an event but a log?

pvital (commented):
Is this only happening on the OTel branch, or should it also be adapted to the master branch?

GSVarsha:
> Is this only happening on the OTel branch, or should it also be adapted to the master branch?

No, this happens only on OTel branch because this change was done by us during migration.

src/instana/span/registered_span.py
GSVarsha: The backend expects the entry to be `span.data["log"]` so when we supply it as `span.data["event"]` it doesn’t recognise the same.

CagriYonca (approved):

pvital (approved):
It looks good to me.

PR #568, opened 2024-07-29
fix: Ensure trace level from current context is considered

Ferenc-:


pvital (approved):
It looks good to me.

src/instana/propagators/http_propagator.py
```diff
@@ -20,6 +20,10 @@ def __init__(self):
     def inject(self, span_context, carrier, disable_w3c_trace_context=False):
         trace_id = span_context.trace_id
         span_id = span_context.span_id
+        # Suppression `level` made in the child context or in the parent context
+        # has priority over any non-suppressed `level` setting
+        child_level = int(self.extract_instana_headers(carrier)[2] or "1")
```
GSVarsha, line 25: Hi @Ferenc- , according to [PEP 333](https://github.com/python/peps/blob/main/peps/pep-0333.rst?plain=1#L720-L722), `response_headers` are a list of tuples. I tested with wsgi frameworks `bottle` and `cherrypy` and both of them produce errors with our existing code on the master branch.

```
2024-08-07 18:38:40,399: 43705 DEBUG instana: extract error:
  File "/Users/varsha/Documents/GitHub/instana-repos/tracer-repos/python-sensor/src/instana/propagators/base_propagator.py", line 221, in extract_instana_headers
    trace_id = dc.get(self.LC_HEADER_KEY_T) or dc.get(self.ALT_LC_HEADER_KEY_T) or dc.get(
AttributeError: 'list' object has no attribute 'get'
```

```
# bottle
<class 'list'> [('Content-Length', '12'), ('Content-Type', 'text/html; charset=UTF-8')]
# cherrypy
<class 'list'> [('Content-Type', 'text/html;charset=utf-8'), ('Server', 'CherryPy/18.10.0'), ('Date', 'Wed, 07 Aug 2024 13:08:40 GMT'), ('Content-Length', '11')]
```

PR #570, opened 2024-07-31
[OTel] urllib3 instrumentation

pvital:


src/instana/instrumentation/urllib3.py
```diff
@@ -2,109 +2,137 @@
...
+                kvs["method"] = kwargs.get("method")
+                kvs["path"] = (
+                    kwargs.get("path") if kwargs.get("path") else kwargs.get("url")
+                )
 
             # Strip any secrets from potential query params
-            if kvs.get('path') is not None and ('?' in kvs['path']):
-                parts = kvs['path'].split('?')
-                kvs['path'] = parts[0]
+            if kvs.get("path") is not None and ("?" in kvs["path"]):
```
GSVarsha, line 59:
```suggestion
            if kvs.get("path") and ("?" in kvs["path"]):
```

pvital: Good catch! ;-)

CagriYonca (approved):
Looks fine to me.

src/instana/instrumentation/urllib3.py
```diff
@@ -2,109 +2,137 @@
...
+                        parts[1],
+                        agent.options.secrets_matcher,
+                        agent.options.secrets_list,
+                    )
+
+            if isinstance(instance, urllib3.connectionpool.HTTPSConnectionPool):
+                kvs["url"] = "https://%s:%d%s" % (kvs["host"], kvs["port"], kvs["path"])
             else:
-                kvs['url'] = 'http://%s:%d%s' % (kvs['host'], kvs['port'], kvs['path'])
+                kvs["url"] = "http://%s:%d%s" % (kvs["host"], kvs["port"], kvs["path"])
```
GSVarsha, line 72: For better readability, can we use f strings here?

src/instana/instrumentation/urllib3.py
```diff
@@ -2,109 +2,137 @@
...
         # If we're not tracing, just return; boto3 has it's own visibility
-        if (tracing_is_off() or (operation_name == 'boto3')):
+        if tracing_is_off() or (span_name == "boto3"):
             return wrapped(*args, **kwargs)
 
-        with tracer.start_active_span("urllib3", child_of=parent_span) as scope:
+        parent_context = parent_span.get_span_context() if parent_span else None
+
+        with tracer.start_as_current_span(
+            "urllib3", span_context=parent_context, kind=SpanKind.CLIENT
```
GSVarsha, line 113: As discussed yesterday, can we remove this?

```suggestion
            "urllib3", span_context=parent_context
```

src/instana/instrumentation/urllib3.py
pvital: I had to work around this once the linter was not getting issues with the f-string, but the runtime was. It will be a bit different.

tests/clients/test_urllib3.py
```diff
@@ -1,501 +1,530 @@
...
-          sleep(10)
-          return x.status_code
+        assert len(spans) == 5
+        nums = map(lambda s: s.data["http"]["params"].split("=")[1], spans)
+        assert set(nums) == set(("1", "2", "3", "4", "5"))
+
+    @pytest.mark.skipif(
+        sys.platform == "darwin",
+        reason="Avoiding ConnectionError when calling multi processes of Flask app.",
+    )
```
GSVarsha, line 58: Can we avoid the import of `pytest` module by using `unittest` to skip on condition?

```suggestion
    @unittest.skipIf(
        sys.platform == "darwin",
        reason="Avoiding ConnectionError when calling multi processes of Flask app.",
    )
```

GSVarsha:
Can you mention the coverage percentage as well?

GSVarsha:
Sorry, accidentally closed with comment. Reopened now.

tests/clients/test_urllib3.py
pvital: 🤔 
As the idea is to remove the dependency of the old `unittest` module, I guess it should be interesting to do it another way. I will make the test class no longer dependent on the `unittest` module. 
What do you think @GSVarsha?

pvital:
> Can you mention the coverage percentage as well?

It's 86% at the moment. The Exception cases are missing. I will work on those and update the PR.

pvital:
> > Can you mention the coverage percentage as well?
> 
> It's 86% at the moment. The Exception cases are missing. I will work on those and update the PR.

It's 100% now.

tests/clients/test_urllib3.py
```diff
@@ -1,501 +1,575 @@
...
-        self.recorder = tracer.recorder
+        self.recorder = tracer.span_processor
         self.recorder.clear_spans()
 
-    def tearDown(self):
-        """ Ensure that allow_exit_as_root has the default value """
+    @pytest.fixture(autouse=True)
+    def _teardown(self) -> None:
+        """Ensure that allow_exit_as_root has the default value"""
         agent.options.allow_exit_as_root = False
```
GSVarsha, line 39: I tried running them and below is the order of their execution
```
tests/clients/test_urllib3.py::TestUrllib3::test_vanilla_requests 
setup
teardown
testcase
```

`yield` keyword or `request.addfinalizer` method is required to indicate the order of setup and teardown methods. Although I read somewhere that `yield` is more preferred.

pvital: > `yield` keyword or `request.addfinalizer` method is required to indicate the order of setup and teardown methods. Although I read somewhere that `yield` is more preferred.

In that way we can have only one method that performs everything, like:
```
@pytest.fixture(autouse=True)
def _setup(self) -> None:
    # Clear all spans before a test run
    self.http = urllib3.PoolManager()
    self.recorder = tracer.span_processor
    self.recorder.clear_spans()
    yield
    # teardown
    # Ensure that allow_exit_as_root has the default value
```

I will make the change.

GSVarsha (approved):
Looks good to me!

PR #572, opened 2024-08-01
[OTel] enable instrumentation tests

pvital:
Add logic to enable the instrumentation tests as their migration is finalized. 

In addition, formated the `__init__.py` file and fixed lint violations, using `ruff` (vscode) to:
- Black-compatible code formatting.
- fix all auto-fixable violations, like unused imports.
- isort-compatible import sorting.

### DIFF
--- a/src/instana/__init__.py
+++ b/src/instana/__init__.py
@@ -1,38 +1,36 @@
 # coding=utf-8
+# (c) Copyright IBM Corp. 2021
+# (c) Copyright Instana Inc. 2016
 """
-▀████▀███▄   ▀███▀▄█▀▀▀█▄███▀▀██▀▀███     ██     ▀███▄   ▀███▀     ██
-  ██   ███▄    █ ▄██    ▀█▀   ██   ▀█    ▄██▄      ███▄    █      ▄██▄
-  ██   █ ███   █ ▀███▄        ██        ▄█▀██▄     █ ███   █     ▄█▀██▄
-  ██   █  ▀██▄ █   ▀█████▄    ██       ▄█  ▀██     █  ▀██▄ █    ▄█  ▀██
-  ██   █   ▀██▄█ ▄     ▀██    ██       ████████    █   ▀██▄█    ████████
-  ██   █     ███ ██     ██    ██      █▀      ██   █     ███   █▀      ██
-▄████▄███▄    ██ █▀█████▀   ▄████▄  ▄███▄   ▄████▄███▄    ██ ▄███▄   ▄████▄
+Instana
 
-https://www.instana.com/
+https://www.ibm.com/products/instana
 
-Documentation: https://www.instana.com/docs/
+Documentation: https://www.ibm.com/docs/en/instana-observability/current
 Source Code: https://github.com/instana/python-sensor
 """
 
-
+import importlib
 import os
 import sys
-import importlib
 
-from .version import VERSION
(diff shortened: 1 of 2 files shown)


pvital:
This PR depends on PR #570.

src/instana/__init__.py
```diff
@@ -1,38 +1,36 @@
...
+    is_webhook_instrumented,
+)
+from instana.version import VERSION
+
+__author__ = "Instana Inc."
+__copyright__ = "Copyright 2020 Instana Inc."
+__credits__ = ["Pavlo Baron", "Peter Giacomo Lombardo", "Andrey Slotin"]
+__license__ = "MIT"
+__maintainer__ = "Peter Giacomo Lombardo"
+__email__ = "peter.lombardo@instana.com"
```
GSVarsha, line 28: I think we should think about changing all the entries except `__license__` 😅

src/instana/__init__.py
```diff
@@ -115,81 +137,91 @@ def lambda_handler(event, context):
...
         # Now get the function and execute it
         if hasattr(handler_module, function_name):
             handler_function = getattr(handler_module, function_name)
             return handler_function(event, context)
         else:
-            print("Couldn't determine and locate default function handler: %s.%s" % (module_name, function_name))
+            print(
+                "Couldn't determine and locate default function handler: %s.%s"
+                % (module_name, function_name)
+            )
```
GSVarsha, line 153: Can we use f-strings?

src/instana/__init__.py
```diff
@@ -115,81 +137,91 @@ def lambda_handler(event, context):
...
+        # from instana.instrumentation import pymysql  # noqa: F401
+        # from instana.instrumentation import psycopg2  # noqa: F401
+        # from instana.instrumentation import redis  # noqa: F401
+        # from instana.instrumentation import sqlalchemy  # noqa: F401
+        from instana.instrumentation import (
+            fastapi_inst,  # noqa: F401
+            flask,  # noqa: F401
+            logging,  # noqa: F401
+            urllib3,  # noqa: F401
+        )
```
GSVarsha, line 196:
```suggestion
        from instana.instrumentation import (
            flask,  # noqa: F401
            logging,  # noqa: F401
            urllib3,  # noqa: F401
        )
```

src/instana/__init__.py
```diff
@@ -115,81 +137,91 @@ def lambda_handler(event, context):
...
+        and len(sys.argv) > 0
+        and (os.path.basename(sys.argv[0]) in do_not_load_list)
+    ):
         if "INSTANA_DEBUG" in os.environ:
-            print("Instana: No use in monitoring this process type (%s).  "
-                  "Will go sit in a corner quietly." % os.path.basename(sys.argv[0]))
+            print(
+                "Instana: No use in monitoring this process type (%s).  "
+                "Will go sit in a corner quietly." % os.path.basename(sys.argv[0])
+            )
```
GSVarsha, line 216:
```suggestion
            print(
                f"Instana: No use in monitoring this process type ({os.path.basename(sys.argv[0])}. Will go sit in a corner quietly."
            )```

src/instana/__init__.py
pvital: We should also do that, but first, we need to investigate internally how to proceed in this case. 
For now, I'd like to keep it as is.

GSVarsha (approved):
Looks good to me!

CagriYonca (approved):

PR #578, opened 2024-08-08
report(currency): google-cloud library versions has been updated

CagriYonca:
Versions of modules below has been updated for the Python runtime versions 3.8, 3.9, 3.10, 3.11, 3.12:
- google-cloud-pubsub version has been changed up to 2.23.0 (JIRA/INSTA-5312)
- google-cloud-storage version has been changed up to 2.14.0 (JIRA/INSTA-5313)

Since grpcio is not supported by beta versions of Python 3.13, we don't support Python 3.13.


pvital:
>     * google-cloud-pubsub version has been changed up to 2.14.0 
>     * google-cloud-storage version has been changed up to 2.23.0 

I guess you changed the order of the versions:

* google-cloud-pubsub's most recent version is 2.23.0.
* google-cloud-storage, you are setting an upper-bound as 2.1.40. I can see the [most recent version](https://pypi.org/project/google-cloud-storage/#history) is 2.18.2. Are the versions between 2.14.0 and 2.18.2 not working?

tests/requirements-310.txt
```diff
@@ -8,8 +8,8 @@ fastapi>=0.92.0
 flask>=2.3.2
 markupsafe>=2.1.0
 grpcio>=1.37.1
-google-cloud-pubsub<=2.1.0
-google-cloud-storage>=1.24.0
+google-cloud-pubsub<=2.23.0
```
pvital, line 11: If we can support the latest version, set a lower-bound version instead of an upper-bound limit.  That way, we can always test and guarantee the most updated version.

```suggestion
google-cloud-pubsub>=2.1.0
```

Please, adapt all other files for this change.

tests/requirements-313.txt
```diff
@@ -15,7 +15,7 @@ markupsafe>=2.1.0
 # Depends on grpcio
 #google-cloud-pubsub<=2.1.0
 #google-cloud-storage>=1.24.0
-lxml>=4.9.2
+# lxml>=4.9.2
```
pvital, line 18: What's the reason for this comment here?

pvital (changes requested):

tests/requirements-313.txt
CagriYonca: After I commented out this module from requirements-313.txt file, it works fine, otherwise pip installation is failing in the container
image

tests/requirements-310.txt
CagriYonca: Sure, updating now.

CagriYonca:
> > ```
> > * google-cloud-pubsub version has been changed up to 2.14.0 
> > * google-cloud-storage version has been changed up to 2.23.0 
> > ```
> 
> I guess you changed the order of the versions:
> 
> * google-cloud-pubsub's most recent version is 2.23.0.
> * google-cloud-storage, you are setting an upper-bound as 2.1.40. I can see the [most recent version](https://pypi.org/project/google-cloud-storage/#history) is 2.18.2. Are the versions between 2.14.0 and 2.18.2 not working?

Yes, I have mistyped version numbers, we're supporting pubsub up to 2.23.0 and storage up to 2.14.0. It's failing for google-cloud-storage versions after 2.14.0:
image

tests/requirements-313.txt
pvital: OK.

tests/requirements-310.txt
```diff
@@ -8,8 +8,8 @@ fastapi>=0.92.0
 flask>=2.3.2
 markupsafe>=2.1.0
 grpcio>=1.37.1
-google-cloud-pubsub<=2.1.0
-google-cloud-storage>=1.24.0
+google-cloud-pubsub>=2.23.0
```
pvital, line 11: You should keep the version number as before and only modify the operation. 
```suggestion
google-cloud-pubsub>=2.1.0
```

pvital (changes requested):

tests/requirements-310.txt
CagriYonca: I'll change it but one question, it was <= 2.1.0 before and now we're changing it as >= 2.1.0, so we'll lose support for the < 2.1.0 . Is that okay?

CagriYonca: I think we should make it <= 2.23.0 because our code is compatible up to 2.23 and may not be compatible with 2.24.0

pvital: Then, our CI pipeline, which checks the versions, will create a new currency ticket every time a new version is released.

Setting a lower-bound version will always get the latest version during our daily tests. If the tests pass, then we will be good for that version. If the tests break, we investigate the reason and fix our code if necessary, or we decide to set the upper-bound version.

Let's go for the lower-bound for now.

tests/requirements-313.txt
Ferenc-: I don't see this change justified just yet. The error message posted above has nothing to do with `lxml`. That is simply a consequence of setting `export AUTOWRAPT_BOOTSTRAP='instana'` before `instana` is even installed.

Ferenc- (changes requested):

tests/requirements-313.txt
pvital: @Ferenc-, I'm sorry, but I didn't get your point. This error is coming from the `pip` install before any test execution. How will the setting of `AUTOWRAPT_BOOTSTRAP` impact this phase?

CagriYonca: I've reverted the change,  it's working on CircleCI. Probably problem was occurring because of arm64 architecture.

Ferenc- (approved):
LGTM

pvital (approved):
It looks good to me.

PR #579, opened 2024-08-09
report(currency): updated starlette version

CagriYonca:
- Added Starlette version 0.38.2 for Python 3.8, 3.9, 3.10, 3.11 and 3.12 has been updated. Python 3.13 has no support for `greenlet` yet which is required for Starlette.
- Added dependency to tests/conftest.py

https://github.com/python-greenlet/greenlet/issues/392

### DIFF
--- a/tests/requirements-gevent-starlette.txt
+++ b/tests/requirements-gevent-starlette.txt
@@ -4,6 +4,6 @@ gevent>=1.4.0
 mock>=2.0.0
 pyramid>=2.0.1
 pytest>=4.6
-starlette>=0.12.13
+starlette>=0.38.2
 urllib3>=1.26.5
 uvicorn>=0.13.4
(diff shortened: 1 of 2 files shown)


tests/requirements-gevent-starlette.txt
```diff
@@ -4,6 +4,6 @@ gevent>=1.4.0
 mock>=2.0.0
 pyramid>=2.0.1
 pytest>=4.6
-starlette>=0.12.13
+starlette>=0.38.2
```
pvital, line 7: This change is unnecessary if the CI pipeline that uses this requirements file works.

pvital (changes requested):

tests/requirements-gevent-starlette.txt
CagriYonca: Done.

PR #589, opened 2024-08-19
Fix: context propagation in nested spans

GSVarsha:

Fix: context propagation in nested spans

### DIFF
--- a/src/instana/tracer.py
+++ b/src/instana/tracer.py
@@ -30,7 +30,7 @@ from instana.propagators.text_propagator import TextPropagator
 from instana.recorder import StanRecorder
 from instana.sampling import InstanaSampler, Sampler
 from instana.span.kind import EXIT_SPANS
-from instana.span.span import InstanaSpan
+from instana.span.span import InstanaSpan, get_current_span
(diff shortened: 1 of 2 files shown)


src/instana/tracer.py
```diff
@@ -118,7 +118,7 @@ def start_span(
         record_exception: bool = True,
         set_status_on_exception: bool = True,
     ) -> InstanaSpan:
-        parent_context = span_context
+        parent_context = get_current_span().get_span_context() if span_context is None else span_context
```
pvital, line 121: I think it's cleaner and easier to read this way. What do you think @GSVarsha ?

```suggestion
        parent_context = span_context if span_context else get_current_span().get_span_context()
```

pvital (changes requested):
Nice catch @GSVarsha. Just a simple request.

src/instana/tracer.py
GSVarsha: Done!

pvital (approved):
It looks good to me.

CagriYonca (approved):
LGTM

PR #590, opened 2024-08-20
[OTel] Refactor Django instrumentation

GSVarsha:

coverage: 95%


tests/apps/app_django.py
```diff
@@ -103,23 +109,22 @@ def not_found(request):
...
-        pscope.span.set_tag(ext.COMPONENT, "Python simple example app")
-        pscope.span.set_tag(ext.SPAN_KIND, ext.SPAN_KIND_RPC_SERVER)
-        pscope.span.set_tag(ext.PEER_HOSTNAME, "localhost")
-        pscope.span.set_tag(ext.HTTP_URL, "/python/simple/one")
-        pscope.span.set_tag(ext.HTTP_METHOD, "GET")
-        pscope.span.set_tag(ext.HTTP_STATUS_CODE, 200)
-        pscope.span.log_kv({"foo": "bar"})
+    with tracer.start_as_current_span("asteroid") as pspan:
+        pspan.set_attribute("component", "Python simple example app")
+        pspan.set_attribute("span.kind", "client")
```
pvital, line 114: Let's keep as a pattern the usage of OTel's SpanKind class enumaration. Please adapt the rest of the code to use like this:

```suggestion
from opentelemetry.trace import SpanKind
        pspan.set_attribute("span.kind", SpanKind.CLIENT)
```

tests/frameworks/test_django.py
```diff
@@ -4,109 +4,114 @@
...
+    def _resource(self) -> Generator[None, None, None]:
         """ Clear all spans before a test run """
-        self.recorder = tracer.recorder
-        self.recorder.clear_spans()
         self.http = urllib3.PoolManager()
+        self.recorder = tracer.span_processor
+        self.recorder.clear_spans()
 
-    def tearDown(self):
+    def tearDown(self) -> None:
```
pvital, line 28: Is this method being used anywhere? If not, you can remove it or move it to the `_resource()` one.

tests/frameworks/test_django.py
```diff
@@ -4,109 +4,114 @@
...
-    def test_basic_request(self):
-        with tracer.start_active_span('test'):
+    def test_basic_request(self) -> None:
+        with tracer.start_as_current_span("test"):
             response = self.http.request('GET', self.live_server_url + '/', fields={"test": 1})
 
-        self.assertTrue(response)
-        self.assertEqual(200, response.status)
+        assert response
+        assert 200 == response.status
```
pvital, line 37: It's not critical, but I suggest you always put the expected value on the right side of the statement. That way, if you want to change this value, it will be easier to modify, and it will also be easier to read the code with them in that part.
```suggestion
        assert response.status == 200
```

pvital (changes requested):
Few requests. The last one you can ignore if you want.

tests/apps/app_django.py
GSVarsha: According to our [instana spec](https://github.ibm.com/instana/technical-documentation/blob/master/tracing/specification/README.md?plain=1#L446), `span.data.sdk.type` must be one of `"entry"`, `"exit"` or `"intermediate"` and in our [codebase](https://github.com/instana/python-sensor/blob/otel_migration/src/instana/span/sdk_span.py#L24), we use `span.kind` to set this attribute.

tests/frameworks/test_django.py
GSVarsha: Yeah I did that for flask tests when I manually changed all of them which was very time consuming so now I'm using vscode's find and replace and regex to change assert statements based on unittest to pytest which keeps the values in the same order as before.

tests/frameworks/test_django.py
GSVarsha: Moved it to `_resource()`!

tests/apps/app_django.py
GSVarsha: Made the required changes as discussed on the call.

pvital (approved):
It looks good to me.

PR #592, opened 2024-08-21
[OTel] Refactor of FastAPI instrumentation.

pvital:
Coverage: 100%


CagriYonca (approved):
Looks good to me!

tests/frameworks/test_fastapi.py
```diff
@@ -1,628 +1,586 @@
...
+        self.recorder.clear_spans()
+
+        # Hack together a manual custom headers list; We'll use this in tests
+        agent.options.extra_http_headers = [
+            "X-Capture-This",
+            "X-Capture-That",
+            "X-Capture-This-Too",
+            "X-Capture-That-Too",
+        ]
+        yield
```
GSVarsha, line 33: Since we don't have a teardown logic here and `autouse` is set to `True`, I think we can safely remove the keyword `yield` here.
```suggestion
```

pvital: Done

GSVarsha (approved):
Looks good to me!

PR #594, opened 2024-08-22
update: added google-cloud-storage v2.18.2 support

CagriYonca:
- added google-cloud-storage v2.18.2 support for python 3.8, 3.9 and 3.10.

Python 3.11 and 3.12 has the dependency conflict below:
image


pvital:
[Python 3.8](https://devguide.python.org/versions/#versions) will be declared EOL in about two months when 3.13 is released. I prefer to find a way to also support it on 3.11 and 3.12 since these will be used more and more for stable apps. 

What happens if we specify on the `requirements.txt` file for 3.11 and 3.12 the `google-api-core >= 2.15.0`?

pvital (commented):
[Python 3.8](https://devguide.python.org/versions/#versions) will be declared EOL in about two months when 3.13 is released. I prefer to find a way to also support it on 3.11 and 3.12 since these will be used more and more for stable apps. 

What happens if we specify on the `requirements.txt` file for 3.11 and 3.12 the `google-api-core >= 2.15.0`?

pvital:
Wonderful @CagriYonca! 

Please change the following in addition:
1. remove any entry of the Google Cloud packages from the requirements file for Python 3.11, 3.12, and 3.13.
2. add an extra job for Python 3.13.
3. update the pipelines for Tekton as well.

CagriYonca:
> Wonderful @CagriYonca!
> 
> Please change the following in addition:
> 
> 1. remove any entry of the Google Cloud packages from the requirements file for Python 3.11, 3.12, and 3.13.
> 2. add an extra job for Python 3.13.
> 3. update the pipelines for Tekton as well.

- Removed Google Cloud packages from reqs file for 3.11, 3.12 and 3.13.
- Python 3.13 is already not supported due to grpcio dependency.
- Tekton pipeline is updated.

pvital (approved):
It looks good to me.

PR #596, opened 2024-08-26
Fix: test_non_async_simple

GSVarsha:
Even with the code before suppression logic caused exception which wasn't logged. Now it fails as expected (due to no support for `httpx` used by `TestClient`)

```
tests/frameworks/test_fastapi.py::TestFastAPI::test_non_async_simple '<' not supported between instances of 'int' and 'NoneType'
```


pvital (approved):
It looks good to me.

PR #597, opened 2024-08-26
[OTel] Refactor of psycopg2 instrumentation

CagriYonca:
- Coverage 96%

PR #598, opened 2024-08-26
[OTel] Refactor of pymysql implementation

CagriYonca:
- 100% coverage

### DIFF
--- a/src/instana/instrumentation/pymysql.py
+++ b/src/instana/instrumentation/pymysql.py
@@ -2,17 +2,17 @@
 # (c) Copyright Instana Inc. 2019
 
 
-from ..log import logger
-from .pep0249 import ConnectionFactory
+from instana.log import logger
+from instana.instrumentation.pep0249 import ConnectionFactory
 
 try:
-    import pymysql #
+    import pymysql
 
-    cf = ConnectionFactory(connect_func=pymysql.connect, module_name='mysql')
+    cf = ConnectionFactory(connect_func=pymysql.connect, module_name="mysql")
 
-    setattr(pymysql, 'connect', cf)
-    if hasattr(pymysql, 'Connect'):
-        setattr(pymysql, 'Connect', cf)
+    setattr(pymysql, "connect", cf)
+    if hasattr(pymysql, "Connect"):
+        setattr(pymysql, "Connect", cf)
 
     logger.debug("Instrumenting pymysql")
 except ImportError:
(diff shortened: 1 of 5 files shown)


src/instana/span/registered_span.py
GSVarsha, line 1: I could see the same changes in psycopg2 PR. Please remove the changes from one of them and merge the PR with the changes first.

tests/clients/test_pymysql.py
GSVarsha, line 1: Please apply the suggestions on [this](https://github.com/instana/python-sensor/pull/597/files#diff-5651d72da7b16df0956b21756082a5245e596b92731d325e455cc647d25ec487) file here as well.

tests/clients/test_pymysql.py
```diff
@@ -39,303 +40,314 @@ def setUp(self):
...
-        with tracer.start_active_span("test"):
+        assert "test" == test_span.data["sdk"]["name"]
+        assert test_span.t == db_span.t
+        assert db_span.p == test_span.s
+        assert 2 == db_span.ec
+
+        assert (
+            db_span.data["mysql"]["error"]
+            == "(1146, \"Table '%s.blah' doesn't exist\")" % testenv["mysql_db"]
+        )
```
GSVarsha, line 249: You can use an f-string here instead of the old syntax of `%` operator.

GSVarsha (changes requested):
Few requests.

src/instana/instrumentation/pymysql.py
```diff
@@ -2,17 +2,17 @@
...
 
 
-from ..log import logger
-from .pep0249 import ConnectionFactory
+from instana.log import logger
+from instana.instrumentation.pep0249 import ConnectionFactory
 
 try:
-    import pymysql #
+    import pymysql  #
```
pvital, line 9: Why is there a lonely `#` at the end of this line?

tests/clients/test_pymysql.py
```diff
@@ -39,303 +37,314 @@ def setUp(self):
...
         if self.db and self.db.open:
-          self.db.close()
+            self.db.close()
         agent.options.allow_exit_as_root = False
 
-    def test_vanilla_query(self):
+    def test_vanilla_query(self) -> None:
         affected_rows = self.cursor.execute("""SELECT * from users""")
-        self.assertEqual(1, affected_rows)
+        assert 1 == affected_rows
```
pvital, line 56: Always put the expected value on the right side of the assert statement. That way, it will be easier to change if necessary in the future and better to read.

Make this change all through this file, please.

```suggestion
        assert affected_rows == 1
```

pvital (changes requested):
A couple of change requests.

pvital:
One more thing, are you not forgetting to remove the following line from the `tests/conftest.py`file?

`collect_ignore_glob.append("*clients/test_pym*")`

pvital (approved):
It looks good to me.

PR #603, opened 2024-08-29
[OTel] unittest coverage improvement

CagriYonca:
I worked on the modules below, increased existing unittests and created non-existing:
- src/instana/collector/base.py (100% coverage)
- src/instana/collector/helpers/runtime.py (97%)
- src/instana/collector/host.py (96%)
- src/instana/collector/utils.py (100%)
- src/instana/propagators/base_propagator.py (96%)
- src/instana/sampling.py (100%)
- src/instana/util/traceutils.py (100%)


src/instana/propagators/base_propagator.py
```diff
@@ -232,15 +251,17 @@ def __determine_span_context(
...
 
         if traceparent:
             ctx_traceparent = traceparent
             ctx_tracestate = tracestate
 
         return SpanContext(
-            trace_id=ctx_trace_id,
-            span_id=ctx_span_id,
+            trace_id=int(ctx_trace_id),
+            span_id=int(ctx_span_id),
```
GSVarsha, line 264: Why is this change needed?

CagriYonca: Because span context is waiting for 2 int values for trace id and span id, and since we use get_instana_ancestor to provide these values, they may be strings (as you can see in the test case below)

image

pvital: Don't change that at this point. 
If you read the code, `ctx_trace_id` and `ctx_span_id` have been set to integers since the beginning (see lines 196 and 197).
If, as you mentioned, the values coming from the `get_instana_ancestor()` are string, then you have to change them by evaluating their type where `ctx_trace_id` and `ctx_span_id`  are overwritten on lines 227 and 228: 
```
if instana_ancestor:
    ctx_trace_id = int(instana_ancestor.t)
    ctx_span_id = int(instana_ancestor.p)
```

CagriYonca: get_traceparent_fields is also returning string below, we'll need to cast to integer it's values as well, you can see the tp_trace_id and tp_parent_id are being set to ctx_trace_id and ctx_span_id in disable_traceparent block:


 ```
  _, tp_trace_id, tp_parent_id, _ = self._tp.get_traceparent_fields(
        traceparent
    )

    if tracestate and "in=" in tracestate:
        instana_ancestor = self._ts.get_instana_ancestor(tracestate)

    if disable_traceparent == "":
        ctx_trace_id = tp_trace_id[-16:]
        ctx_span_id = tp_parent_id
        ctx_synthetic = synthetic
        ctx_trace_parent = True
        ctx_instana_ancestor = instana_ancestor
        ctx_long_trace_id = tp_trace_id
    else:
        if instana_ancestor:
            ctx_trace_id = int(instana_ancestor.t)
            ctx_span_id = int(instana_ancestor.p)
            ctx_synthetic = synthetic

```

Comment (pvital) [reply src/instana/propagators/base_propagator.py:264]: > get_traceparent_fields is also returning string below, we'll need to cast to integer it's values as well, you can see the tp_trace_id and tp_parent_id are being set to ctx_trace_id and ctx_span_id in disable_traceparent block:

So, go ahead and do it there as well! 😉

Comment (CagriYonca) [reply src/instana/propagators/base_propagator.py]: Done 👍

File: src/instana/collector/base.py [hunk 2]
```diff
@@ -150,12 +150,6 @@ def background_report(self):
 
         self.prepare_and_report_data()
 
-        if self.thread_shutdown.is_set():
-            logger.debug(
-                "Thread shutdown signal is active: Shutting down reporting thread"
-            )
-            return False
-
```
GSVarsha, line 158: Any specific reason to remove this?

src/instana/span_context.py
```diff
@@ -1,6 +1,7 @@
 # (c) Copyright IBM Corp. 2021
 # (c) Copyright Instana Inc. 2019
 
+from opentelemetry.trace import SpanContext as OtelSpanContext
```
GSVarsha, line 4: Import already available [here](https://github.com/instana/python-sensor/pull/603/files#diff-56b8f1677bf62524a185ea8a6e852056cfadb96a983b0270772abb92cb39c5cfR8).

src/instana/span.py
GSVarsha, line 1: This file is outdated. Please take a look at `src/instana/span` folder if you want to make some changes.

GSVarsha (changes requested):
Few requests ⬆️

src/instana/span.py
pvital: 👆🏻 this!

tests/collector/__init__.py
pvital, line 1: You are moving/renaming the `__init__.py` file from the old `platforms` directory to the `collector`, and letting the new `agents` directory without an init file.

I suggest you to:
- First, rename the new `agents` dir name to `agent` - coherent with the same name in `src/instana/` 
- Second, create a new `__init__.py` file in the `tests/collector` folder and move this one to the new `agent` dir.

tests/util/test_traceutils.py
```diff
@@ -0,0 +1,65 @@
...
+    tracing_is_off,
+)
+from instana.singletons import agent, tracer
+
+
+class TestTraceUtils:
+    @pytest.fixture(autouse=True)
+    def _resource(self):
+        pass
+
```
pvital, line 18: No need to have this if you are not setting up and tearing down the test execution.

tests/util/test_traceutils.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_extract_custom_headers(self, span) -> None:
```
pvital, line 19: Missing type annotation.

pvital (changes requested):
Few requests

src/instana/collector/base.py
CagriYonca: It was removed by a mistake, re-added it

PR #606, opened 2024-09-02
Fix OTel logging tests

pvital:
- fix: Adapt logging tests to OTel after rebase.
- tests(logging): Refactor to pure pytest UT.


GSVarsha (approved):
Looks good to me.

PR #607, opened 2024-09-02
[OTel] Refactor of Pymongo insrumentation

CagriYonca:
Coverage: 91%

### DIFF
--- a/src/instana/span/registered_span.py
+++ b/src/instana/span/registered_span.py
@@ -252,14 +252,8 @@ class RegisteredSpan(BaseSpan):
             self.data["pg"]["error"] = span.attributes.pop("pg.error", None)
 
         elif span.name == "mongo":
-            service = "%s:%s" % (
-                span.attributes.pop("host", None),
-                span.attributes.pop("port", None),
-            )
-            namespace = "%s.%s" % (
-                span.attributes.pop("db", "?"),
-                span.attributes.pop("collection", "?"),
-            )
+            service = f"{span.attributes.pop(SpanAttributes.SERVER_ADDRESS, None)}:{span.attributes.pop(SpanAttributes.SERVER_PORT, None)}"
+            namespace = f"{span.attributes.pop(SpanAttributes.DB_NAME, '?')}.{span.attributes.pop(SpanAttributes.DB_MONGODB_COLLECTION, '?')}"
 
             self.data["mongo"]["service"] = service
             self.data["mongo"]["namespace"] = namespace
--- a/tests/clients/test_pymongo.py
+++ b/tests/clients/test_pymongo.py
@@ -2,256 +2,289 @@
 # (c) Copyright Instana Inc. 2020
 
 import json
-import unittest
 import logging
+from typing import Generator
 
-from ..helpers import testenv
-from instana.singletons import agent, tracer
-
-import pymongo
 import bson
+import pymongo
+import pytest
+
+from instana.singletons import agent, tracer
+from instana.span.span import get_current_span
+from tests.helpers import testenv
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


src/instana/span/registered_span.py
```diff
@@ -252,17 +253,20 @@ def _populate_exit_span_data(self, span) -> None:
...
-                span.attributes.pop("host", None),
-                span.attributes.pop("port", None),
+                span.attributes.pop(SpanAttributes.SERVER_ADDRESS, None),
+                span.attributes.pop(SpanAttributes.SERVER_PORT, None),
             )
             namespace = "%s.%s" % (
-                span.attributes.pop("db", "?"),
-                span.attributes.pop("collection", "?"),
+                span.attributes.pop(SpanAttributes.DB_NAME, "?"),
+                span.attributes.pop(SpanAttributes.DB_MONGODB_COLLECTION, "?"),
```
pvital, line 261: As commented [here](https://github.com/instana/python-sensor/pull/604#discussion_r1741233140), we have to check if the backend can understand `db.mongodb.collection` as a valid key instead of `collection`. 

For that, you can run an E2E test with the `master` branch and check the values on the UI. Then, you run the same test with this feature branch and compare all displayed values.

tests/span/test_readable_span.py
```diff
@@ -1,91 +1,109 @@
...
-    assert span.end_time == end_time
-    assert span.attributes == attributes
-    assert span.events == events
-    assert span.status == status
-    assert span.duration == end_time - start_time
-    assert span.stack == stack
+class TestEvent:
+    @pytest.fixture(autouse=True)
+    def _resource(self) -> Generator[None, None, None]:
+        self.test_event = None
```
pvital, line 14: I don't see the necessity to have test_event as a class variable. Since you create the' Event' objects directly in each test method, you can remove this fixture and test the `test_event` there.

tests/span/test_readable_span.py
```diff
@@ -1,91 +1,109 @@
...
+        self,
+        span_context: SpanContext,
+        trace_id: int,
+        span_id: int,
+    ) -> None:
+        span_name = "test-span"
+        timestamp = time.time_ns()
+        self.span = ReadableSpan(span_name, span_context)
+
+        assert self.span is not None
```
pvital, line 54:
```suggestion
        assert self.span
```

tests/span/test_readable_span.py
```diff
@@ -1,91 +1,109 @@
...
+        assert isinstance(self.span, ReadableSpan)
+        assert self.span.name == span_name
+
+        span_context = self.span.context
+        assert isinstance(span_context, SpanContext)
+        assert span_context.trace_id == trace_id
+        assert span_context.span_id == span_id
```
pvital, line 61: Instead of overwriting the `span_context` you are passing as argument, this suggestion would be better.
```suggestion
        assert self.span.context == span_context
        assert isinstance(self.span.context, SpanContext)
        assert self.span.context.trace_id == trace_id
        assert self.span.context.span_id == span_id
```

tests/clients/test_pymongo.py
```diff
@@ -2,10 +2,16 @@
 # (c) Copyright Instana Inc. 2020
 
 import json
+from typing import Generator
 import unittest
```
pvital, line 6: This import is not necessary.

pvital (changes requested):

src/instana/span/registered_span.py
CagriYonca: Created a [new task on JIRA](https://jsw.ibm.com/browse/INSTA-13881) regarding backend development, reverted changes for now.

pvital (approved):
It looks good to me.

PR #609, opened 2024-09-04
fix: psycopg2 merge fix

CagriYonca:
- Removed boto3 from ignored folders list
- Updated one test case due to version update of instana

image


pvital (approved):
It looks good to me.

GSVarsha (approved):
LGTM

PR #612, opened 2024-09-09
fix: Remove an unsupported Span data field.

pvital:
Removed the data field "l" (which stores the span level) from the Span data since the Instana Backend does not support it.


CagriYonca (approved):
LGTM

GSVarsha (approved):
Looks good to me!

PR #622, opened 2024-09-19
refactor: Remove TestAgent from the production code.

pvital:
Remove the TestAgent and the if-blocks that check if running in a test environment from the production code.

**FINALLY**


GSVarsha (approved):
Looks good to me! ✨

PR #627, opened 2024-09-24
[OTel] Guarantee propagators tests after OTel migration.

pvital:
- Fixed propagators and traceparent files to work after OTel migration, following the OTel and W3C specifications.
- TODO later:
  - Adapted propagators and w3c files to follow new style guidelines.
  - Coverage will be handled later.


GSVarsha (approved):
Looks good to me.

PR #628, opened 2024-09-25
[OTel] Grpcio instrumentation

GSVarsha:
`coverage: 95%`
- Type annotations will be handled later

### DIFF
--- a/src/instana/instrumentation/grpcio.py
+++ b/src/instana/instrumentation/grpcio.py
@@ -1,27 +1,32 @@
 # (c) Copyright IBM Corp. 2021
 # (c) Copyright Instana Inc. 2019
 
-
-import wrapt
-import opentracing
-
-from ..log import logger
-from ..singletons import tracer
-
 try:
     import grpc
-    from grpc._channel import _UnaryUnaryMultiCallable, _StreamUnaryMultiCallable, \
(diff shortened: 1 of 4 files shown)


src/instana/instrumentation/grpcio.py
```diff
@@ -33,248 +38,308 @@ def collect_tags(span, instance, argv, kwargs):
...
+                    disable_w3c_trace_context=True,
+                )
+                collect_attributes(span, instance, argv, kwargs)
+                span.set_attribute("rpc.call_type", "unary")
 
                 rv = wrapped(*argv, **kwargs)
             except Exception as e:
-                scope.span.log_exception(e)
+                span.record_exception(e)
                 raise
```
pvital, line 82: You can skip this `except` block from the coverage (that it's unnecessary to check) and reduce the number of error spans logged if you use the following:

```
    except Exception as exc:
        span.record_exception(exc)
else:
```

You can apply this for the rest of the code (and other files).

pvital (changes requested):
One request only.

src/instana/instrumentation/grpcio.py
GSVarsha: Done.

pvital (approved):
It looks good to me.

PR #630, opened 2024-09-25
fix: Remove an unsupported top level span attribute.

GSVarsha:
Related to https://github.com/instana/python-sensor/pull/612.
- Update the respective tests.


pvital (approved):
It looks good to me.

PR #635, opened 2024-09-29
[OTel] Fix `sanic` headers capture, `couchbase` imports, `sdk.custom.tags`

GSVarsha:
- fix(sanic): capture headers only if `agent.options.extra_http_headers` is `True`
- fix(sanic): skip instrumentation for unsupported versions
- fix(couchbase): fix skipping instrumentation for unsupported versions. Earlier, it would fail even if `couchbase` does not have `__version__` attribute.
- minor fix(couchbase): skip importing other modules if the package to be instrumented is not present
- fix: capture custom tags on SDK spans: https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#span-data


CagriYonca (approved):
LGTM

pvital (approved):
It looks good to me.

PR #636, opened 2024-09-29
[OTel] Fix: test_span caplog setup

CagriYonca:
This PR is needed for Celery instrumentation, solves the test_span unittest errors
- Added logger configurations to catch Status class logs using caplog 
- Made some format enhancement and typing annotations
- Restructured unittests using TestSpan class


GSVarsha (approved):
LGTM

PR #637, opened 2024-09-30
[OTel] quick fix

pvital:
- fix: Span duration and start_time when recording.
Following the Instana Tracing Specification, the span `ts` must express the span's start as milliseconds since the Unix epoch, and the span `d` must be the duration of the span in milliseconds.

- fix(tests): Skipping MySQL Client on MacOS.


CagriYonca (approved):
Looks good to me, goodbye MySQL errors 🚀

GSVarsha:
> Span duration and start_time when recording.
Following the Instana Tracing Specification, the span ts must express the span's start as milliseconds since the Unix epoch, and the span d must be the duration of the span in milliseconds.

But that's why we are converting them from nano seconds to milli seconds, right?

pvital:
> Looks good to me, goodbye MySQL errors 🚀

Sorry, @CagriYonca, but it will be there for a while.
I read the wrong line of the documentation, and the calculation is correct, as mentioned by @GSVarsha.

I am canceling this PR.

PR #640, opened 2024-10-03
unittest: formatted test folder structure and improved unittest coverage

CagriYonca:
I worked on the modules below, increased existing unittests and created non-existing:
•⁠  ⁠src/instana/collector/base.py (98% coverage)
•⁠  ⁠src/instana/collector/helpers/runtime.py (97%)
•⁠  ⁠src/instana/collector/host.py (96%)
•⁠  ⁠src/instana/collector/utils.py (100%)
•⁠  ⁠src/instana/propagators/base_propagator.py (96%)
•⁠  ⁠src/instana/sampling.py (93%)
•⁠  ⁠src/instana/util/traceutils.py (97%)

Modified folder structures in tests folder, used the same structure that src/instana folder has.


tests/conftest.py
```diff
@@ -142,6 +142,8 @@ def always_true(_: object) -> bool:
 @pytest.fixture(autouse=True)
 def can_send(monkeypatch, request) -> None:
     """Return always True for HostAgent.can_send()"""
+    if "no_fixture" in request.node.keywords:
+        return None
```
pvital, line 146: What is the reason for this if-block? I did not understand it.

tests/platforms/test_host.py
pvital, line 1: Could you transform this into a `pytest` file instead of a `unittest`?

In addition, you can also remove the `create_agent_and_setup_tracer()` as explained before.

tests/span/test_base_span.py
```diff
@@ -1,158 +1,182 @@
...
-    base_span = BaseSpan(span, None)
-    value = span
-
-    converted_value = base_span._convert_attribute_value(value)
-    assert "<instana.span.span.InstanaSpan object" in converted_value
+class TestBaseSpan:
+    @pytest.fixture(autouse=True)
+    def _resource(self) -> Generator[None, None, None]:
+        self.base_span = None
+        yield
```
pvital, line 19: If this is the only change in this file, I don't see any reason to create a class for the whole test. One of the advantages of using pytest is that it is unnecessary to create a test class to run tests.

tests/span/test_event.py
```diff
@@ -1,36 +1,63 @@
...
-    assert "field1" in event.attributes.keys()
-    assert "two" == event.attributes.get("field2")
-    assert event.timestamp == timestamp
+class TestEvent:
+    @pytest.fixture(autouse=True)
+    def _resource(
+        self,
+    ) -> Generator[None, None, None]:
+        self.event = None
+        yield
```
pvital, line 17: The same as before. If this is the only change in this file, I don't see any reason to create a class for the whole test. One of the advantages of using pytest is that it is unnecessary to create a test class to run tests.

tests/util/test_traceutils.py
```diff
@@ -0,0 +1,69 @@
...
+    get_tracer_tuple,
+    tracing_is_off,
+)
+
+
+class TestTraceUtils:
+    @pytest.fixture(autouse=True)
+    def _resource(self) -> Generator[None, None, None]:
+        pass
+
```
pvital, line 22: Again, it is unnecessary to create a test class to run tests.

pvital (changes requested):
few requests

tests/agent/test_host.py
```diff
@@ -1,310 +1,630 @@
...
+        assert isinstance(self.agent.options, StandardOptions)
+
+    def test_agent_default_log_level(self) -> None:
+        self.create_agent_and_setup_tracer()
+        assert self.agent.options.log_level == logging.WARNING
+
+    def test_agent_instana_debug(self) -> None:
+        os.environ["INSTANA_DEBUG"] = "asdf"
+        self.create_agent_and_setup_tracer()
+        assert self.agent.options.log_level == logging.DEBUG
```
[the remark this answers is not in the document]
CagriYonca: What do you think about this?

```suggestion
        self.agent.options = StandardOptions()
        assert self.agent.options.log_level == logging.DEBUG
```

tests/conftest.py
CagriYonca: For some of test cases I didn't want to use fixture, such as below, with this if block, I'm being able to skip fixtures:

```
    @pytest.mark.no_fixture
    def test_can_send(
        self,
    ) -> None:
        agent = HostAgent()
        agent._boot_pid = 12345
        with patch.object(os, "getpid", return_value=12344), patch(
            "instana.agent.host.HostAgent.handle_fork"
        ) as mock_handle, patch.dict("os.environ", {}, clear=True):
            agent.can_send()
            assert agent._boot_pid == 12344
            mock_handle.assert_called_once()

            with patch.object(agent.machine.fsm, "current", "wait4init"):
                assert agent.can_send() is True
```

tests/span/test_base_span.py
CagriYonca: Isn't having the same structure in all test files a good behaviour?

pvital: IMHO, only when necessary. That's not the case here.

pvital (approved):
It looks good to me.

PR #644, opened 2024-10-04
Fix OTel header inject values.

pvital:
### Fix OTel header inject values (Part 1)
Instana's Tracer Specification claims the `X-INSTANA-T` and `X-INSTANA-S ` values injected in returned headers must be 16 lowercase hexadecimal characters.

#### Before:
```
  2 passing (11s)
  42 failing
```
#### After:
```
  14 passing (30s)
  30 failing
```

### DIFF
--- a/tests/frameworks/test_grpcio.py
+++ b/tests/frameworks/test_grpcio.py
@@ -10,7 +10,7 @@ import grpc
 
 from opentelemetry.trace import SpanKind
 
-import tests.apps.grpc_server
+import tests.apps.grpc_server  # noqa: F401
 import tests.apps.grpc_server.stan_pb2 as stan_pb2
 import tests.apps.grpc_server.stan_pb2_grpc as stan_pb2_grpc
 from tests.helpers import testenv, get_first_span_by_name
@@ -29,11 +29,12 @@ class TestGRPCIO:
         self.server_stub = stan_pb2_grpc.StanStub(self.channel)
         # The grpc client apparently needs a second to connect and initialize
         time.sleep(1)
+        yield
         # tearDown
         # Ensure that allow_exit_as_root has the default value
         agent.options.allow_exit_as_root = False
 
-    def generate_questions(self) -> None:
+    def generate_questions(self) -> Generator[None, None, None]:
         """Used in the streaming grpc tests"""
         questions = [
             stan_pb2.QuestionRequest(question="Are you there?"),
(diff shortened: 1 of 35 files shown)


CagriYonca (approved):
LGTM

tests/frameworks/test_grpcio.py
```diff
@@ -690,6 +691,7 @@ def test_root_exit_span(self) -> None:
         assert not server_span.data["rpc"]["error"]
 
     def test_no_root_exit_span(self) -> None:
+        agent.options.allow_exit_as_root = True
```
GSVarsha, line 694: Please remove this line or change it to False. We already have a testcase to check for root exit span when set to True. In this testcase, we check if no spans are recorded when set to False.

```suggestion
        agent.options.allow_exit_as_root = False
```

pvital: Curiously, the missing of this setting was causing one flaky test error.
I changed the code.

tests/clients/test_logging.py
```diff
@@ -79,7 +79,7 @@ def test_parameters(self) -> None:
         assert spans[0].data["log"].get("parameters") is not None
 
     def test_no_root_exit_span(self) -> None:
-        agent.options.allow_exit_as_root = True
+        agent.options.allow_exit_as_root = False
```
GSVarsha, line 82: I understand the confusion caused by the same testcase name as `grpcio` 😅 but the usecase here is to not record the span even though `agent.options.allow_exit_as_root` is set to `True` since its an `INFO` log.

GSVarsha (changes requested):
We might have to adapt `tracestate` and `Server-Timing` headers as well.

```
% curl -D- http://127.0.0.1:8000/      
HTTP/1.1 200 OK
Connection: keep-alive
Keep-Alive: 5
X-INSTANA-L: 1
traceparent: 00-000000000000000074477b5a47051399-74477b5a47051399-01
tracestate: in=8378801259414885273;8378801259414885273
X-INSTANA-T: 74477b5a47051399
X-INSTANA-S: 74477b5a47051399
Server-Timing: intid;desc=8378801259414885273
Content-Length: 55
Content-Type: application/json

{"msg":"Hello world. Running on sanic version: 19.9.0"}
```

tests/clients/test_logging.py
pvital: OK. As a suggestion, it would be good to add additional comments for different behaviors or make explicit what the goal of the tests is in the method name.

pvital:
> We might have to adapt `tracestate` and `Server-Timing` headers as well.

You are right. That will demand some extra work since almost all header settings are made in each instrumentation code. I will update the PR and let you know.

GSVarsha (approved):
LGTM

PR #643, opened 2024-10-04
fix(ci): flaky tests

GSVarsha:
### Fix flaky tests
- Always return `True` for `HostAgent.is_agent_listening()`
- Register pytest marker: `pytest.mark.original`
- Fix warnings

#### CircleCI
Screenshot 2024-10-04 at 3 19 18 PM

#### Tekton CI
Screenshot 2024-10-04 at 3 34 26 PM

PR #646, opened 2024-10-07
Fix: Flaky tests (Part 2)

GSVarsha:
Part 2 of https://github.com/instana/python-sensor/pull/643.


pvital (approved):
It looks good to me.

PR #647, opened 2024-10-08
Fix(currency): Run only starlette tests in `gevent_starlette` task

GSVarsha:
Run only starlette tests for now in `gevent_starlette` task

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -327,8 +327,10 @@ jobs:
       - pip-install-deps:
           requirements: "tests/requirements-gevent-starlette.txt"
       - run-tests-with-coverage-report:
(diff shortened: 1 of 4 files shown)


pvital (changes requested):
You can keep the environment variable's name as `GEVENT_STARLETTE_TEST` to explicitly show that this will run both Starlette and Gevent tests in a separate job and let only line 41 of the `conftest.py` file be commented.

With that, you can continue running Starlette in a separate job, temporarily prevent Gevent from running, and remove the Starlette tests from the runtime jobs.

pvital (approved):
It looks good to me.

CagriYonca (approved):
LGTM

PR #648, opened 2024-10-10
fix: _log() Multiple Values for 'stacklevel'.

pvital:
The error `_log() got multiple values for keyword argument 'stacklevel'` may occur when you pass both a positional argument and a keyword argument with the same name (`stacklevel`) to the `Logger._log()` function.

This fix checks if the `stacklevel` is a key of the `kwargs` dictionary before calling the `Logger._log()` function, considering it if present.


GSVarsha (approved):
Looks good to me!

CagriYonca (approved):
LGTM

PR #652, opened 2024-10-16
Backport from `legacy_2.x` - fix: _log() Multiple Values for 'stacklevel'.

pvital:
This is a backport from the `legacy_2.x` branch.

The error `_log() got multiple values for keyword argument 'stacklevel'` may occur when you pass both a positional argument and a keyword argument with the same name (`stacklevel`) to the `Logger._log()` function.

This fix checks if the `stacklevel` is a key of the `kwargs` dictionary before calling the `Logger._log()` function, considering it if present.

(cherry picked from commit aaa03d4a51fb78d9b0029784d9792c2e36337082)

### DIFF
--- a/src/instana/instrumentation/logging.py
+++ b/src/instana/instrumentation/logging.py
@@ -22,16 +22,22 @@ def log_with_instana(
     # argv[0] = level
     # argv[1] = message
     # argv[2] = args for message
-    if sys.version_info >= (3, 13):
-        stacklevel = 3
+
+    # We take into consideration if `stacklevel` is already present in `kwargs`.
+    # This prevents the error `_log() got multiple values for keyword argument 'stacklevel'`
+    if "stacklevel" in kwargs.keys():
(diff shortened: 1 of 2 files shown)


GSVarsha (changes requested):
Hello @pvital, if the user tries to do something like [this](https://github.com/instana/python-sensor/compare/fix_logging_multiple_stacklevel...adapt_logging_incoming_stacklevel#diff-0cbfa5cea993c2341e4074f56f45edd74ae67d3ee333e7c0fa2e715f2c219cf8), then our existing code might not give the desired result. So I think it will be better to do something like [this](https://github.com/instana/python-sensor/compare/fix_logging_multiple_stacklevel...adapt_logging_incoming_stacklevel#diff-0801e6e45225c6850147ad2b53d790cfccb7c21d3f78d4d422c18d1d30c9f774). 

What do you think?

PR #655, opened 2024-10-24
fix: provide ASGI support with Django

GSVarsha:
Fix: Provide ASGI support with Django.

#### Why?
`request.environ` is available only with requests of type `WSGIRequest` and not `ASGIRequest`.
[`request.headers`](https://docs.djangoproject.com/en/3.0/ref/request-response/#django.http.HttpRequest.headers) is only available from `django-2.2`.

Since we provide support for `django>=1.11` we'll use [`request.META`](https://docs.djangoproject.com/en/5.1/ref/request-response/#django.http.HttpRequest.META).

### DIFF
--- a/src/instana/instrumentation/django/middleware.py
+++ b/src/instana/instrumentation/django/middleware.py
@@ -2,7 +2,6 @@
 # (c) Copyright Instana Inc. 2018
 
 
-import os
 import sys
 
 import opentracing as ot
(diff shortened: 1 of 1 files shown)


pvital (approved):
It looks good to me.

CagriYonca (approved):
Looks good to me!

arjun-rajappa (approved):

PR #660, opened 2024-10-28
Adapt to logger's incoming stacklevel

GSVarsha:
Related to https://github.com/instana/python-sensor/pull/652

Bypass the stack level of our instrumentation when the keyword `stacklevel` is set as well.

### Before the changes <sup>py3.11</sup>
log_with_instana -> log_custom_warning -> main
`stacklevel1 -> stacklevel2 -> stacklevel3`

### After the changes <sup>py3.11</sup>
log_custom_warning -> main -> test_log_caller_with_stacklevel
`stacklevel1 -> stacklevel2 -> stacklevel3`

PR #661, opened 2024-10-28
fix: Adapt to logger's incoming stacklevel

GSVarsha:
Related to https://github.com/instana/python-sensor/pull/652

Bypass the stack level of our instrumentation when the keyword `stacklevel` is set as well.

More details on https://github.com/instana/python-sensor/pull/660.


pvital (commented):
It is not necessary to backport your solution for the legacy branch @GSVarsha.
The customer has already tested the previous patch and it runs with no issues. 
Thanks anyway.

GSVarsha:
> It is not necessary to backport your solution for the legacy branch @GSVarsha. The customer has already tested the previous patch and it runs with no issues. Thanks anyway.

I will close this issue then :)

PR #667, opened 2024-11-05
enhancement: removed protobuf from requirements list

CagriYonca:
This PR removes the protobuf from dependencies list to solve the problem below version 4.0.0:

```
AttributeError: module 'google._upb._message' has no attribute 'Message'
TODO: Remove this when support for 4.0.0 is done
protobuf<4.0.0
```
Updating protobuf is also required to solve google-cloud-storage dependency problems for versions >= 2.14.2 for python 3.11 and 3.12


pvital (approved):
It looks good to me.
Good job @CagriYonca.

PR #669, opened 2024-11-06
Fix(flaky tests): added announce function to conftest

CagriYonca:
- Added announce function to conftest
- Removed duplicated code file


tests/agent/test_host.py
```diff
@@ -64,7 +64,7 @@ def test_has_options(self) -> None:
         assert isinstance(self.agent.options, StandardOptions)
 
     def test_agent_default_log_level(self) -> None:
-        assert self.agent.options.log_level == logging.WARNING
+        assert self.agent.options.log_level == logging.WARN
```
GSVarsha, line 67:
```suggestion
        assert self.agent.options.log_level == logging.WARNING
```
Why was this change made? I can see that they are essentially the same.
```
WARNING = 30
WARN = WARNING
```

PR #675, opened 2024-11-27
ci: Add new pipeline to test Python 3.14.0a2.

pvital:
Add the new alpha2 test build for Python 3.14.0 after being announced [here](https://www.python.org/downloads/release/python-3140a2/).


GSVarsha (approved):
Looks good to me!

PR #677, opened 2024-12-12
ci: updated pipeline to filter src, tests and tests_aws folders

CagriYonca:
Updated circleci pipeline to provide filtering src, tests and tests_aws folders before running the pipeline.
Added exception for nightly builds of the main


arjun-rajappa (approved):
LGTM!

PR #682, opened 2025-01-14
ci: Automate package release and publish.

pvital:
This commit adds automation to release and publish into GitHub and PyPI.org any new package version based on the creation of a new tag.


GSVarsha (approved):
Looks good to me!

PR #686, opened 2025-01-16
Fix: Add new AWS region `ap-southeast-7` to publish

GSVarsha:
Add new AWS region `ap-southeast-7` to publish.


pvital (approved):
It looks good to me.

PR #688, opened 2025-01-17
[FUP Optimisation] redis: added endpoint filtering

CagriYonca:
Added endpoint filtering for redis instrumentation. We've aligned with Java and NodeJS teams. The customer has three options to filter endpoints:
- Defining ignore-endpoint through the configuration.yaml file of the agent
- Using environment-variable before initialisation
- Using src/instana/configurator.py

If the customer used more than one way of these three, we'd be using one according to this priority:
- Environment variable > in-code setting > agent configuration

## Ignoring endpoints through env variable
The customer can use filtering with the environment variable below. If the env variable is set, other two ways will be ignored.

- `INSTANA_IGNORE_ENDPOINTS="service:endpoint"`
- `INSTANA_IGNORE_ENDPOINTS="service:endpoint1,endpoint2"`
- Filtering redis instrumentation at all:
  `INSTANA_IGNORE_ENDPOINTS="service"`
- Filtering multiple instrumentations:
  `INSTANA_IGNORE_ENDPOINTS="service1:endpoint1;service2:endpoint1"`
  `INSTANA_IGNORE_ENDPOINTS="service1;service2"`

## Ignoring endpoints through the configurator.py file
The customer can add a configuration line like below:

- `config["tracing"]["ignore_endpoints"] = {"service1": ["endpoint1", "endpoint2"], "service2": []}`

## Ignoring endpoints through the agent configuration
The customer can add a configuration to agent-folder/etc/instana/configuration.yaml

``` yaml
com.instana.tracing:
  ignore-endpoints:
    service1:    # Add package name, e.g., redis, dynamodb
      - 'endpoint1' # Add endpoint
      - 'endpoint2'
    service2:
      - 'endpoint1'
```

Reference: https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#ignoring-endpoints

### DIFF
--- a/src/instana/agent/host.py
+++ b/src/instana/agent/host.py
@@ -6,25 +6,30 @@ The in-process Instana agent (for host based processes) that manages
 monitoring state and reporting that data.
 """
 
-import os
 import json
+import os
 from datetime import datetime
+from typing import Any, Dict, List, Optional, Union
+from requests import Response
 
-import urllib3
 import requests
+import urllib3
 
-from ..log import logger
-from .base import BaseAgent
-from ..fsm import TheMachine
-from ..version import VERSION
-from ..options import StandardOptions
-from ..collector.host import HostCollector
-from ..util import to_json
-from ..util.runtime import get_py_source
+from instana.agent.base import BaseAgent
+from instana.collector.host import HostCollector
+from instana.configurator import config
+from instana.fsm import Discovery, TheMachine
+from instana.log import logger
+from instana.options import StandardOptions
+from instana.util import to_json
+from instana.util.config import parse_ignored_endpoints
+from instana.util.runtime import get_py_source
+from instana.version import VERSION
 
 
 class AnnounceData(object):
-    """ The Announce Payload """
+    """The Announce Payload"""
+
     pid = 0
     agentUuid = ""
 
@@ -38,10 +43,11 @@ class HostAgent(BaseAgent):
     parts it handles are the announce state and the collection and reporting of metrics and spans to the
     Instana Host agent.
     """
+
     AGENT_DISCOVERY_PATH = "com.instana.plugin.python.discovery"
     AGENT_DATA_PATH = "com.instana.plugin.python.%d"
 
-    def __init__(self):
+    def __init__(self) -> None:
         super(HostAgent, self).__init__()
 
         self.announce_data = None
@@ -54,12 +60,14 @@ class HostAgent(BaseAgent):
         # Update log level from what Options detected
         self.update_log_level()
 
-        logger.info("Stan is on the scene.  Starting Instana instrumentation version: %s", VERSION)
+        logger.info(
+            f"Stan is on the scene.  Starting Instana instrumentation version: {VERSION}"
+        )
 
         self.collector = HostCollector(self)
         self.machine = TheMachine(self)
 
-    def start(self):
+    def start(self) -> None:
         """
         Starts the agent and required threads
 
@@ -68,14 +76,14 @@ class HostAgent(BaseAgent):
         logger.debug("Starting Host Collector")
         self.collector.start()
 
-    def handle_fork(self):
+    def handle_fork(self) -> None:
         """
         Forks happen.  Here we handle them.
         """
         # Reset the Agent
         self.reset()
 
-    def reset(self):
+    def reset(self) -> None:
         """
         This will reset the agent to a fresh unannounced state.
         :return: None
@@ -87,7 +95,7 @@ class HostAgent(BaseAgent):
         # Will schedule a restart of the announce cycle in the future
         self.machine.reset()
 
-    def is_timed_out(self):
+    def is_timed_out(self) -> bool:
         """
         If we haven't heard from the Instana host agent in 60 seconds, this
         method will return True.
@@ -99,7 +107,7 @@ class HostAgent(BaseAgent):
                 return True
         return False
 
-    def can_send(self):
+    def can_send(self) -> bool:
         """
         Are we in a state where we can send data?
         @return: Boolean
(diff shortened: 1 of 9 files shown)


src/instana/collector/helpers/runtime.py
```diff
@@ -10,29 +11,92 @@
...
+                    ignore_endpoints.append(instrumentation)
+
+        # If set from agent configuration (dict format)
+        elif isinstance(params, dict):
+            for service, endpoints in params.items():
+                if not endpoints:  # If endpoints list is empty
+                    ignore_endpoints.append(service)
+                else:
+                    for endpoint in endpoints:
+                        ignore_endpoints.append(f"{service}.{endpoint}")
```
pvital, line 85: Let's practice the `Clean Code` here. It would be better to split this whole function in two: one to parse string ignored endpoints and another to parse dictionary ignored endpoints.

src/instana/instrumentation/redis.py
```diff
@@ -43,6 +43,8 @@ def execute_command_with_instana(
         args: Tuple[object, ...],
         kwargs: Dict[str, Any],
     ) -> object:
+        if check_if_ignored("redis", args[0]):
+            return
```
pvital, line 47: Let's prevent the load and import of all these classes. Move this if block out of the try-except block and all the imports (between lines 8 and 10) to below the `import redis`.

src/instana/options.py
```diff
@@ -13,57 +13,79 @@
...
+                str(os.environ["INSTANA_EXTRA_HTTP_HEADERS"]).lower().split(";")
+            )
+
+        if "INSTANA_IGNORE_ENDPOINTS" in os.environ:
+            self.ignore_endpoints = parse_ignored_endpoints(
+                os.environ["INSTANA_IGNORE_ENDPOINTS"]
+            )
+        else:
+            if "tracing" in config:
+                if "ignore_endpoints" in config["tracing"]:
```
pvital, line 54:
```suggestion
            if "tracing" in config and "ignore_endpoints" in config["tracing"]:
```

src/instana/util/traceutils.py
```diff
@@ -64,4 +92,16 @@ def get_tracer_tuple() -> (
...
+
+def check_if_ignored(
+    instrumentation: str,
+    command: str,
+) -> bool:
+    """Check if the given instrumentation and command combination should be ignored."""
+    return (
+        instrumentation in agent.options.ignore_endpoints
+        or f"{instrumentation}.{command.lower()}" in agent.options.ignore_endpoints
+    )
```
pvital, line 107: A better name for this method is `is_service_or_endpoint_ignored`; the variable names must be coherent.
```suggestion
def is_service_or_endpoint_ignored(
    service: str,
    endpoint: str,
) -> bool:
    """Check if the given service and endpoint combination should be ignored."""
    return (
        service in agent.options.ignore_endpoints
        or f"{service}.{endpoint.lower()}" in agent.options.ignore_endpoints
    )
```

pvital (changes requested):
Few change requests.

src/instana/options.py
CagriYonca: The inner if block is used to prevent errors on misuse of config["tracing"] as you can see below:
```python
>>> config = {"tracing": 15}
>>> "endpoint" in config["tracing"]
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'int' is not iterable
```

src/instana/collector/helpers/runtime.py
```diff
@@ -10,29 +11,92 @@
...
+
+            instrumentation_pairs = params.split(";")
+
+            for pair in instrumentation_pairs:
+                instrumentation = pair.strip()
+
+                if not instrumentation:
+                    continue
+
+                if ":" in pair:
```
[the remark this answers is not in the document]
CagriYonca: Using "in" seems more used by the community:
https://stackoverflow.com/a/27138045/14637536

src/instana/options.py
pvital: I don't see differences from what was suggested:

```python
>>> config = {"tracing": 15}
>>>
>>> "tracing" in config and "ignore_endpoints" in config["tracing"]
  File "<python-input-3>", line 1, in <module>
    "tracing" in config and "ignore_endpoints" in config["tracing"]
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'int' is not iterable
>>>
>>>
>>> config = {"bla": "ble"}
>>> "tracing" in config and "ignore_endpoints" in config["tracing"]
False
>>>
```
If you expect a specific type of value and want to avoid user misconfiguration, you should use `isinstance(config["tracing"], expected_type)` or handle the `TypeError` exception in a try-except block.

src/instana/collector/helpers/runtime.py
```diff
@@ -10,29 +11,115 @@
...
+        if isinstance(params, str):
+            return parse_ignored_endpoints_string(params)
+        elif isinstance(params, dict):
+            return parse_ignored_endpoints_dict(params)
+        else:
+            return []
+    except Exception as e:
+        logger.debug("Error parsing ignored endpoints: %s", str(e))
+        return []
+
```
pvital, line 114: I think these methods are in the wrong place since they are not part of the `Collector` scope but the `Config` one.
It should be located at `src/instana/util/runtime.py` or, even better, in a new file called `src/instana/util/config.py`.

src/instana/options.py
```diff
@@ -13,57 +13,79 @@
     - AWSFargateOptions - Options class for AWS Fargate.  Holds settings specific to AWS Fargate.
     - GCROptions - Options class for Google cloud Run.  Holds settings specific to GCR.
 """
+
 import os
 import logging
+from typing import Any, Dict
+
+from instana.collector.helpers.runtime import parse_ignored_endpoints
```
pvital, line 21: As commented above, the `parse_ignored_endpoints` function should be imported from `src/instana/util/runtime.py` or, even better, from `src/instana/util/config.py`.

tests/collector/helpers/test_collector_runtime.py
```diff
@@ -66,3 +69,32 @@ def test_collect_gc_metrics(self) -> None:
...
+        response = parse_ignored_endpoints(test_pair)
+        assert response == [
+            "service1.endpoint1",
+            "service1.endpoint2",
+            "service2.endpoint1",
+        ]
+
+        test_pair = ""
+        response = parse_ignored_endpoints(test_pair)
+        assert response == []
```
pvital, line 100: If the above suggestions about the methods' location are implemented, could you move these tests to the respective test files?

pvital (changes requested):
Well done, @CagriYonca, but I still have some comments.

pvital (changes requested):
Few more request changes.

As discussed in our call, let's change the approach and filter the spans before sending them to the Agent (or Collector) - see `src/instana/agent/host.py:300`. That way, we can have a one-size-fits-all solution, and no changes to the instrumentation code will be necessary.

src/instana/agent/host.py
```diff
@@ -297,73 +324,147 @@ def report_data_payload(self, payload):
...
         except Exception as exc:
-            logger.debug("report_data_payload: Instana host agent connection error (%s)", type(exc), exc_info=True)
+            logger.debug(
+                f"report_data_payload: Instana host agent connection error ({type(exc)})",
+                exc_info=True,
+            )
         return response
 
-    def handle_agent_tasks(self, task):
+    def report_metrics(self, payload: Dict[str, Any]) -> Union[Response, None]:
```
pvital, line 333: Use `Optional` instead of `Union[None, something]`.
https://docs.python.org/3.13/library/typing.html#typing.Optional 

```suggestion
    def report_metrics(self, payload: Dict[str, Any]) -> Optional[Response]:
```

src/instana/agent/host.py
```diff
@@ -297,73 +324,147 @@ def report_data_payload(self, payload):
...
+        metric_count = len(payload["metrics"])
```
pvital, line 334: It would be better to avoid `KeyError` exceptions here, even if `metrics` is a key that always will be available. I suggest using the `get()` method since this will always be a dictionary.

```python
>>> payload = {"spans": ["ha", "he"]}
>>> metric_count = len(payload["metrics"])
  File "<python-input-7>", line 1, in <module>
    metric_count = len(payload["metrics"])
                       ~~~~~~~^^^^^^^^^^^
KeyError: 'metrics'
>>> metric_count = len(payload.get("metrics", []))
>>> metric_count
0
>>>
```

 ```suggestion
        metric_count = len(payload.get("metrics", []))
```

File: src/instana/agent/host.py [hunk 11]
```diff
@@ -297,73 +324,147 @@ def report_data_payload(self, payload):
...
+                self.__data_url(),
+                data=to_json(metric_bundle),
+                headers={"Content-Type": "application/json"},
+                timeout=0.8,
+            )
+            return response
+        return
+
+    def report_profiles(self, payload: Dict[str, Any]) -> Union[Response, None]:
+        profile_count = len(payload["profiles"])
```
pvital, line 347: Same changes as suggested for `report_metrics()`.

src/instana/agent/host.py
```diff
@@ -297,73 +324,147 @@ def report_data_payload(self, payload):
...
+                data=to_json(payload["profiles"]),
+                headers={"Content-Type": "application/json"},
+                timeout=0.8,
+            )
+            return response
+        return
+
+    def report_spans(self, payload: Dict[str, Any]) -> Union[Response, None]:
+        filtered_spans = self.filter_spans(payload["spans"])
+        if filtered_spans:
```
pvital, line 361:
```suggestion
    def report_spans(self, payload: Dict[str, Any]) -> Optional[Response]:
        filtered_spans = self.filter_spans(payload.get("spans", []))
        if len(filtered_spans) > 0:
```

pvital (changes requested):
Few requests.

pvital (approved):
It looks good to me.

PR #690, opened 2025-01-20
fix: use the correct image for 3.14.0a4

GSVarsha:
### Before
```
Configuration is 'default' on 3.14.0a3 with dependencies in 'requirements-pre314.txt'
```

### After
```
Configuration is 'default' on 3.14.0a4 with dependencies in 'requirements-pre314.txt'
```


pvital (approved):
It looks good to me.

PR #692, opened 2025-02-06
chore(version): Bump version to 3.3.0

CagriYonca:
chore(version): Bump version to 3.3.0

### DIFF
--- a/src/instana/version.py
+++ b/src/instana/version.py
@@ -3,4 +3,4 @@
 
 # Module version file.  Used by setup.py and snapshot reporting.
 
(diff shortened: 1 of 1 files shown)


src/instana/version.py
```diff
@@ -3,4 +3,4 @@
 
 # Module version file.  Used by setup.py and snapshot reporting.
 
-VERSION = "3.2.1"
+VERSION = "3.2.2"
```
pvital, line 6: This new version releases new features, so it must be `3.3.0`. 
```suggestion
VERSION = "3.3.0"
```

PR #693, opened 2025-02-07
fix: capturing extra-http-headers from agent config

CagriYonca:
Added extra-http-headers support to tracer's agent configuration headers.

https://www.ibm.com/docs/en/instana-observability/current?topic=cha-configuring-host-agents-by-using-agent-configuration-file#capturing-custom-http-headers

### DIFF
--- a/src/instana/agent/host.py
+++ b/src/instana/agent/host.py
@@ -17,12 +17,10 @@ import urllib3
 
 from instana.agent.base import BaseAgent
 from instana.collector.host import HostCollector
-from instana.configurator import config
 from instana.fsm import Discovery, TheMachine
 from instana.log import logger
 from instana.options import StandardOptions
 from instana.util import to_json
-from instana.util.config import parse_ignored_endpoints
 from instana.util.runtime import get_py_source
 from instana.version import VERSION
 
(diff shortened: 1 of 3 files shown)


src/instana/agent/host.py
pvital, line 1: Good catch, @CagriYonca, but instead of having these as methods from the `HostAgent` class, the correct place would be in the `BaseOptions` class as they set or modify information regarding the options, not the Agent. 

With that, we can do something like the following code in line 177:

```python
        if "secrets" in res_data:
            self.options.set_secrets(res_data["secrets"])
        if "tracing" in res_data:
            self.options.set_tracing(res_data["tracing"])
        else:
            if "extraHeaders" in res_data:
                self.options.set_extra_headers(res_data["extraHeaders"])
```

pvital (changes requested):
One request.

pvital (changes requested):
Few grammar corrections, but I am missing unit tests for all these new functions.

Remember, it is a best practice to create unit tests for every new function, even if it's a single-line one. This way, we can guarantee full coverage of our code.

pvital (approved):
It looks good to me.  
Great job @CagriYonca!

PR #695, opened 2025-02-13
fix: remove unsupported span attribute

GSVarsha:
Remove unsupported redundant span attribute `span.data.custom.attributes`

```
{"data":{...,"custom":{"attributes":{"span.kind":"<SpanKind.SERVER: 1>"}}},...}
```


pvital (approved):
That's nice! 
I am working on an improvement that will include the property `kind` as part of the `InstanaSpan`. One of the changes I was making was to remove these attributes. 😄 

It looks good to me.

PR #699, opened 2025-02-19
Add support and instrumentation to Kafka.

pvital:
This PR contains all changes to support Kafka and the instrumentation implementation for the `kafka-python` package.


tests/clients/test_kafka_python.py
```diff
@@ -0,0 +1,120 @@
...
+from kafka.admin import KafkaAdminClient, NewTopic
+from opentelemetry.trace import SpanKind
+
+from instana.singletons import agent, tracer
+from tests.helpers import testenv
+
+
+class TestKafkaPythonProducer:
+    @pytest.fixture(autouse=True)
+    def _setup(self) -> Generator[None, None, None]:
```
CagriYonca, line 16:
```suggestion
    def _resource(self) -> Generator[None, None, None]:
```
We use _resource as the setup function name

src/instana/span/registered_span.py
```diff
@@ -1,11 +1,21 @@
...
+    LOCAL_SPANS,
+)
 
-from opentelemetry.trace import SpanKind
-from opentelemetry.semconv.trace import SpanAttributes
+if TYPE_CHECKING:
+    from instana.span.span import InstanaSpan
 
 
 class RegisteredSpan(BaseSpan):
```
CagriYonca, line 21: Could we add typing hints to the functions we worked on? All missing typing hints will be covered as we create PRs.
The parameters of the `__init__` function could be updated accordingly for this PR.

src/instana/span/registered_span.py
```diff
@@ -34,6 +44,10 @@ def __init__(self, span, source, service_name, **kwargs) -> None:
         if "gcps" in span.name:
             self.n = "gcps"
 
+        # unify the span name for kafka-producer and kafka-consumer
+        if "kafka" in span.name:
```
CagriYonca, line 48: Can we combine the change of L145 with this if block?
```suggestion
        if "kafka" in span.name:
                self.n = "kafka"
                self._collect_kafka_attributes(span)
```

CagriYonca (commented):
A few commands, great work Paulo!

src/instana/span/registered_span.py
pvital: We could, but the methods/functions where `self._collect_kafka_attributes(span)` is called aim to set the `span.data`, not the name.

Let's follow the Clean Code principle that functions must only do one thing. In this case, the functions `_populate_{entry, local, exit}_span_data()`  set the `span.data`.  We can create one specific function to set the span name correctly. 

I plan to refactor this file later to make it cleaner, so I'd like to keep it as is for now.

CagriYonca (approved):

PR #700, opened 2025-02-21
feat: Add instrumentation to confluent-kafka-python.

pvital:


src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+        with tracer.start_as_current_span(
+            "kafka-consumer", span_context=parent_context, kind=SpanKind.CONSUMER
+        ) as span:
+            span.set_attribute("kafka.access", "poll")
+
+            try:
+                res = wrapped(*args, **kwargs)
+                span.set_attribute("kafka.service", res.topic() if res else None)
```
GSVarsha, line 145: Suggesting this change because `kafka.service` will be set to `None` [here](https://github.com/instana/python-sensor/pull/700/files#diff-93ccde7c403c9606145fbbc078ce3338b1f052099504b9c0172bae66564ade5bL373) even if we don't set the value below.
```suggestion
                if res:
                  span.set_attribute("kafka.service", res.topic())
```

pvital: Good catch @GSVarsha. Thanks.

tests/clients/kafka/test_confluent_kafka.py
```diff
@@ -0,0 +1,199 @@
...
+        assert not test_span.ec
+        assert not kafka_span.ec
+
+        assert kafka_span.n == "kafka"
+        assert kafka_span.k == SpanKind.CLIENT
+        assert kafka_span.data["kafka"]["service"] == testenv["kafka_topic"]
+        assert kafka_span.data["kafka"]["access"] == "produce"
+
+    def test_trace_confluent_kafka_consume(self) -> None:
+        agent.options.allow_exit_as_root = False
```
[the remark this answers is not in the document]
pvital: The consumption of Kafka topics can be started in different services, not linked to an HTTP one - it can be a crontab script or a service only listening to the topic in a k8s pod. 

But I will remove it and create specific `exit_as_root` tests later in a different PR.

GSVarsha (approved):
Looks good to me!

PR #706, opened 2025-03-10
feat: Add Spyne instrumentation

GSVarsha:
feat: Add Spyne instrumentation

### DIFF
--- a/src/instana/__init__.py
+++ b/src/instana/__init__.py
@@ -188,6 +188,7 @@ def boot_agent() -> None:
             sqlalchemy,  # noqa: F401
             starlette,  # noqa: F401
             urllib3,  # noqa: F401
+            spyne,
         )
         from instana.instrumentation.aiohttp import (
             client as aiohttp_client,  # noqa: F401
--- a/src/instana/__init__.py
+++ b/src/instana/__init__.py
@@ -188,7 +188,7 @@ def boot_agent() -> None:
             sqlalchemy,  # noqa: F401
             starlette,  # noqa: F401
             urllib3,  # noqa: F401
-            spyne,
+            spyne,  # noqa: F401
         )
         from instana.instrumentation.aiohttp import (
             client as aiohttp_client,  # noqa: F401
--- a/src/instana/instrumentation/spyne.py
+++ b/src/instana/instrumentation/spyne.py
@@ -12,11 +12,37 @@ try:
     from instana.util.secrets import strip_secrets_from_query
     from instana.util.traceutils import extract_custom_headers
 
+    def set_span_attributes(span, headers):
+        if "REQUEST_METHOD" in headers:
+            span.set_attribute(SpanAttributes.HTTP_METHOD, headers["REQUEST_METHOD"])
+        if "PATH_INFO" in headers:
+            span.set_attribute(SpanAttributes.HTTP_URL, headers["PATH_INFO"])
+        if "QUERY_STRING" in headers and len(headers["QUERY_STRING"]):
+            scrubbed_params = strip_secrets_from_query(
+                headers["QUERY_STRING"],
+                agent.options.secrets_matcher,
+                agent.options.secrets_list,
+            )
+            span.set_attribute("http.params", scrubbed_params)
+        if "HTTP_HOST" in headers:
+            span.set_attribute("http.host", headers["HTTP_HOST"])
+
+    def set_response_status_code(span, response_string):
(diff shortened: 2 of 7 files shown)


src/instana/__init__.py
```diff
@@ -188,6 +188,7 @@ def boot_agent() -> None:
             sqlalchemy,  # noqa: F401
             starlette,  # noqa: F401
             urllib3,  # noqa: F401
+            spyne,
```
CagriYonca, line 191: Interesting that it didn't ask for F401

src/instana/instrumentation/spyne.py
```diff
@@ -0,0 +1,130 @@
...
+        span = ctx.udc
+        response_string = ctx.transport.resp_code
+
+        if span and response_string:
+            set_response_status_code(span, response_string)
+            if span.is_recording():
+                span.end()
+
+            ctx.udc = None
+        return wrapped(*args, **kwargs)
```
CagriYonca, line 95: Same as above

src/instana/span/kind.py
```diff
@@ -18,6 +18,7 @@
     "urllib3",
     "wsgi",
     "asgi",
+    "spyne"
```
CagriYonca, line 21: Spyne isn't supported by the backend as I see, did you check it?

CagriYonca (commented):
Left some comments.

src/instana/instrumentation/spyne.py
```diff
@@ -0,0 +1,130 @@
...
+
+            response_headers = ctx.transport.resp_headers
+
+            extract_custom_headers(span, response_headers, format=False)
+            tracer.inject(span.context, Format.HTTP_HEADERS, response_headers)
+
+            response = wrapped(*args, **kwargs)
+
+            set_response_status_code(span, ctx.transport.resp_code)
+            return response
```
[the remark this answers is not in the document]
pvital: @CagriYonca I don't think it's necessary as the errors are captured by the `ctx.transport.resp_code`and handled at `set_response_status_code()`.   

The `try-except` is used to handle the possible errors coming to the execution of the `wrapped()` function, but it looks like `spyne` sets all errors in the `ctx` and  @GSVarsha is elegantly handling in a specific function - super clean code 😄.

src/instana/span/kind.py
pvital: @CagriYonca raises a good point, but as `spyne` is an RPC toolkit, shouldn't we set the span names as `rpc-client` or `rpc-server`?

pvital (commented):
Few comments. Please address @CagriYonca requests.

src/instana/__init__.py
GSVarsha: Usually what happens when we don't add `# noqa: F401`?

src/instana/instrumentation/spyne.py
GSVarsha: Resolving this conversation. More info: https://github.com/instana/python-sensor/pull/706#discussion_r1988585859

src/instana/span/kind.py
GSVarsha: Changed span name to `rpc-server`

src/instana/span/kind.py
```diff
@@ -32,6 +33,7 @@
     "gcps-consumer",
     "asgi",
     "kafka-consumer",
+    "spyne"
```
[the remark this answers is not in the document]
GSVarsha: Changed span name to `rpc-server`

CagriYonca (approved):
Looks good to me, Good Job Varsha!

src/instana/__init__.py
pvital: Nothing happens if you don't have a linter or code formatted configured in your system. 

If you have one, like [ruff](https://docs.astral.sh/ruff/) (which we use in our [pre-commit](https://pre-commit.com/) [configuration](https://github.com/instana/python-sensor/blob/main/.pre-commit-config.yaml)), you should see an error complaining about unused imports ([F401](https://docs.astral.sh/ruff/rules/unused-import/)).

Please keep the pattern for the rest of the code and add the `# noqa: F401` for this line.

pvital (changes requested):
Only one minor request.

src/instana/__init__.py
GSVarsha: Got it! Added `# noqa: F401` to the import.

PR #707, opened 2025-03-10
ft: add fup support for dynamodb

CagriYonca:
- Added Fair Usage Policy support for DynamoDB.

### DIFF
--- a/src/instana/agent/host.py
+++ b/src/instana/agent/host.py
@@ -22,6 +22,7 @@ from instana.log import logger
 from instana.options import StandardOptions
 from instana.util import to_json
 from instana.util.runtime import get_py_source
+from instana.util.span_utils import get_operation_specifier
 from instana.version import VERSION
 
 
@@ -346,14 +347,16 @@ class HostAgent(BaseAgent):
         return
 
     def filter_spans(self, spans: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
-        from instana.util.traceutils import is_service_or_endpoint_ignored
-
+        """
+        Filters given span list using ignore-endpoint variable and returns the list of filtered spans.
+        """
         filtered_spans = []
         for span in spans:
             if (hasattr(span, "n") or hasattr(span, "name")) and hasattr(span, "data"):
                 service = span.n
-                endpoint = span.data[service]["command"]
-                if isinstance(endpoint, str) and is_service_or_endpoint_ignored(
+                operation_specifier = get_operation_specifier(service)
+                endpoint = span.data[service][operation_specifier]
+                if isinstance(endpoint, str) and self.__is_service_or_endpoint_ignored(
                     service, endpoint
                 ):
                     continue
@@ -362,6 +365,16 @@ class HostAgent(BaseAgent):
             else:
                 filtered_spans.append(span)
         return filtered_spans
+
+    def __is_service_or_endpoint_ignored(
+        self, service: str, endpoint: str = ""
+    ) -> bool:
+        """Check if the given service and endpoint combination should be ignored."""
+
+        return (
+            service.lower() in self.options.ignore_endpoints
+            or f"{service.lower()}.{endpoint.lower()}" in self.options.ignore_endpoints
+        )
 
     def handle_agent_tasks(self, task: Dict[str, Any]) -> None:
         """
--- a/src/instana/util/span_utils.py
+++ b/src/instana/util/span_utils.py
@@ -1,0 +1,13 @@
+# (c) Copyright IBM Corp. 2025
+
+from typing import Optional
+
+
+def get_operation_specifier(span_name: str) -> Optional[str]:
+    """Get the specific operation specifier for the given span."""
+    operation_specifier = ""
+    if span_name == "redis":
+        operation_specifier = "command"
+    elif span_name == "dynamodb":
+        operation_specifier = "op"
+    return operation_specifier
--- a/src/instana/util/traceutils.py
+++ b/src/instana/util/traceutils.py
@@ -15,10 +15,10 @@ from typing import (
 from instana.log import logger
 from instana.singletons import agent, tracer
 from instana.span.span import get_current_span
-from instana.tracer import InstanaTracer
 
 if TYPE_CHECKING:
     from instana.span.span import InstanaSpan
+    from instana.tracer import InstanaTracer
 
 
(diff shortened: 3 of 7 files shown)


src/instana/util/traceutils.py
```diff
@@ -104,3 +104,13 @@ def is_service_or_endpoint_ignored(
         service.lower() in agent.options.ignore_endpoints
         or f"{service.lower()}.{endpoint.lower()}" in agent.options.ignore_endpoints
     )
+
+
+def get_operation_specifier(span_name: str) -> str:
+    """Get the specific operation specifier for the given span."""
+    operation_specifier = ""
```
pvital, line 111: I suggest you return `Optional[str]` and handle the `None` return when calling this function.
```suggestion
def get_operation_specifier(span_name: str) -> Optional[str]:
    """Get the specific operation specifier for the given span name."""
    operation_specifier = None
```

tests/util/test_traceutils.py
```diff
@@ -117,3 +118,13 @@ def test_is_service_or_endpoint_ignored() -> None:
...
+
+
+def test_get_operation_specifier() -> None:
+    test_redis = "redis"
+    response_redis = get_operation_specifier(test_redis)
+    assert response_redis == "command"
+
+    test_dynamodb = "dynamodb"
+    response_dynamodb = get_operation_specifier(test_dynamodb)
+    assert response_dynamodb == "op"
```
pvital, line 130: Nice work on this UT. I suggest you use parametrized tests for it; then, we can easily extend any other service we want to use the FUP feature.
```suggestion
@pytest.mark.parametrize(
        "span_name, expected_result",
        [
            ("something", None),
            ("redis", "command"),
            ("dynamodb", "op"),
        ],
    )
def test_get_operation_specifier(span_name: str, expected_result: Optional[str]) -> None:
    response = get_operation_specifier(span_name)
    assert response == expected_result
```

src/instana/agent/host.py
```diff
@@ -346,13 +346,17 @@ def report_spans(self, payload: Dict[str, Any]) -> Optional[Response]:
         return
 
     def filter_spans(self, spans: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
-        from instana.util.traceutils import is_service_or_endpoint_ignored
+        from instana.util.traceutils import (
+            is_service_or_endpoint_ignored,
+            get_operation_specifier,
+        )
```
pvital, line 352: Could you please move these imports to the top of the file?

src/instana/agent/host.py
```diff
@@ -346,13 +346,17 @@ def report_spans(self, payload: Dict[str, Any]) -> Optional[Response]:
         return
 
     def filter_spans(self, spans: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
```
pvital, line 348: Please add a docstring explaining why this method should be used.

pvital (changes requested):
Few requests.

src/instana/agent/host.py
CagriYonca: These imports cause circular imports on top. Could we keep it as is?

pvital: If there are circle import errors, the code is doing something it is not supposed to do or was not well designed. I think you should try to solve them.

CagriYonca: Solved!

src/instana/agent/host.py
```diff
@@ -346,14 +347,17 @@ def report_spans(self, payload: Dict[str, Any]) -> Optional[Response]:
...
-        from instana.util.traceutils import is_service_or_endpoint_ignored
-
+        """
+        Before propagating spans to the agent, this function is being called to filter
+        span list using ignore-endpoint variable.
+        """
```
pvital, line 353: Ok, this is not clear to me. What is this function really doing? 🤷🏻‍♂️
Is the return a list of spans that I should ignore or a list of valid spans to send to the agent?

src/instana/util/span_utils.py
pvital, line 1: Missing copyright header.

src/instana/util/traceutils.py
pvital, line 1: Please don't import inside functions. As we talked about before, try to solve the circular imports by moving the code to the correct place - if there is a circular import, that means the code is not in the proper place.

tests/util/test_traceutils.py
```diff
@@ -103,17 +104,28 @@ def test_is_service_or_endpoint_ignored() -> None:
...
+
+@pytest.mark.parametrize(
+    "span_name, expected_result",
+    [("something", ""), ("redis", "command"), ("dynamodb", "op")],
+)
+def test_get_operation_specifier(
+    span_name: str, expected_result: Optional[str]
+) -> None:
+    response_redis = get_operation_specifier(span_name)
+    assert response_redis == expected_result
```
pvital, line 131: If this function is placed in the `span_utils.py` file, it is placed wrongly here. 
Please create a new test file for the new file created.

pvital (changes requested):
Few comments.

tests/util/test_traceutils.py
```diff
@@ -103,17 +102,17 @@ def test_is_service_or_endpoint_ignored() -> None:
...
 
     # ignore only endpoint1 of service2
-    assert is_service_or_endpoint_ignored("service2", "endpoint1")
-    assert not is_service_or_endpoint_ignored("service2", "endpoint2")
+    assert agent._HostAgent__is_service_or_endpoint_ignored("service2", "endpoint1")
+    assert not agent._HostAgent__is_service_or_endpoint_ignored("service2", "endpoint2")
 
     # don't ignore other services
-    assert not is_service_or_endpoint_ignored("service3")
+    assert not agent._HostAgent__is_service_or_endpoint_ignored("service3")
```
pvital, line 118: If you removed the method `is_service_or_endpoint_ignored()` from the `src/instana/util/traceutils.py` file, this whole test is no longer valid. You can remove it.

Now, since you moved the method `is_service_or_endpoint_ignored()` to the `src/instana/agent/host.py`, you can move this whole function to the `tests/agent/test_host.py` file. BTW, I am missing a test for the method `filter_spans()` from the  `src/instana/agent/host.py`.

pvital (changes requested):
One more little request.

pvital (approved):
It looks good to me.

PR #709, opened 2025-03-13
fix: added exception for pika consumer

CagriYonca:
Test_consume case was failing due to the break inside __consume function. When we break the for loop, generator propagates GeneratorExit exception. I've added an exception case and closed the generator gracefully.


GSVarsha (approved):
Looks apt to me!

PR #715, opened 2025-04-02
fix(sanic): Handle headers extraction after `multidict` update

GSVarsha:
### Overview
- `sanic.compat.Header` contains `__dict__` attribute but does not store the headers dictionary in it. Hence directly convert the `CIMultiDict` instance to dictionary using `dict()`.


CagriYonca (approved):
Good catch!

pvital (approved):
It looks good to me.

PR #716, opened 2025-04-09
ci: downgrade wheel package version

CagriYonca:
Circle CI pipelines were failing for Python 3.9-3.10-3.11 versions due to the last version of `wheels` package. This PR temporarily fixes the issue until a new version of the Wheels package is released.

Main Pipeline
image


pvital (approved):
It looks OK to me.

PR #718, opened 2025-04-10
feat: Kafka FUP Phase I

CagriYonca:
Changes:
- The customers are able to use `configuration.yaml` file to filter specific services and methods now, as specified in [the technical specification](https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#instana_ignore_endpoints_path).
- Specific topics and specific methods can be filtered using one of the below, `configuration.yaml` file has the highest priority, use `send` for kafka-python and `produce` for confluent-kafka:
  - os.environ["INSTANA_IGNORE_ENDPOINTS"] = "kafka:send,consume"
  - os.environ["INSTANA_IGNORE_ENDPOINTS"] = "kafka:produce,consume"
  - os.environ["INSTANA_IGNORE_ENDPOINTS"] = "kafka"
  - os.environ["INSTANA_IGNORE_ENDPOINTS_PATH"] = "configuration.yaml"

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,7 @@ dependencies = [
     "opentelemetry-api>=1.27.0",
     "opentelemetry-semantic-conventions>=0.48b0",
     "typing_extensions>=4.12.2",
+    "pyyaml>=6.0.2",
 ]
 
 [project.entry-points."instana"]
--- a/tests/util/test_config.py
+++ b/tests/util/test_config.py
@@ -1,10 +1,13 @@
+import logging
 from typing import Generator
 
 import pytest
 
 from instana.util.config import (
+    parse_endpoints_of_service,
     parse_ignored_endpoints,
     parse_ignored_endpoints_dict,
+    parse_ignored_endpoints_from_yaml,
     parse_service_pair,
 )
 
@@ -15,28 +18,28 @@ class TestConfig:
         yield
 
     def test_parse_service_pair(self) -> None:
-        test_string = "service1:endpoint1,endpoint2"
+        test_string = "service1:method1,method2"
         response = parse_service_pair(test_string)
-        assert response == ["service1.endpoint1", "service1.endpoint2"]
+        assert response == ["service1.method1", "service1.method2"]
 
         test_string = "service1;service2"
         response = parse_ignored_endpoints(test_string)
-        assert response == ["service1", "service2"]
+        assert response == ["service1.*", "service2.*"]
 
         test_string = "service1"
         response = parse_ignored_endpoints(test_string)
-        assert response == ["service1"]
+        assert response == ["service1.*"]
 
         test_string = ";"
         response = parse_ignored_endpoints(test_string)
         assert response == []
 
-        test_string = "service1:endpoint1,endpoint2;;;service2:endpoint1;;"
+        test_string = "service1:method1,method2;;;service2:method1;;"
         response = parse_ignored_endpoints(test_string)
         assert response == [
-            "service1.endpoint1",
-            "service1.endpoint2",
-            "service2.endpoint1",
+            "service1.method1",
+            "service1.method2",
+            "service2.method1",
         ]
 
         test_string = ""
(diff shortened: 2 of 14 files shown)


tests/util/test_configuration-1.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - query
+    kafka: 
+      - methods: ["consume", "send"]
+        endpoints: ["span-topic", "topic1", "topic2"]
+      - methods: ["consume"]
+        endpoints: ["topic3"]
+      - methods: ["*"] # Applied to all methods
+        endpoints: ["span-topic", "topic4"]
+      # - methods: ["consume", "send"]
+      #   endpoints: ["*"] # Applied to all topics
```
GSVarsha, line 17: Please add a new line at EOF

tests/util/test_configuration-2.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - query
+    kafka: 
+      - methods: ["consume", "send"]
+        endpoints: ["span-topic", "topic1", "topic2"]
+      - methods: ["consume"]
+        endpoints: ["topic3"]
+      - methods: ["*"] # Applied to all methods
+        endpoints: ["span-topic", "topic4"]
+      # - methods: ["consume", "send"]
+      #   endpoints: ["*"] # Applied to all topics
```
GSVarsha, line 17: Same as https://github.com/instana/python-sensor/pull/718/files#r2041487836

pyproject.toml
```diff
@@ -52,6 +52,7 @@ dependencies = [
     "opentelemetry-api>=1.27.0",
     "opentelemetry-semantic-conventions>=0.48b0",
     "typing_extensions>=4.12.2",
+    "pyyaml>=6.0.2",
```
GSVarsha, line 55: Why is `pyyyaml` a dependency for the tracer?

pvital: @GSVarsha, that's because we need to read YAML configuration files now. This is part of the design.

I don't like the idea of having it as one of the main dependencies of our package. Perhaps we can add it as an extra installation option. The users would then execute something like the following command if they want to use the Kafka filtering:

```
pip install instana[yaml]
```

pvital:
@CagriYonca, please work on the #721 first, and after it is merged, we can finalize this one and the following one.

tests/util/test_config.py
pvital, line 1: Missing copyright headers here.

tests/util/test_config.py
```diff
@@ -15,125 +18,150 @@ def _resource(self) -> Generator[None, None, None]:
         yield
```
pvital, line 18: You can remove this `_resource()` function if no pre-setting is necessary.

tests/util/test_config_reader.py
pvital, line 1: Missing copyright headers here.

tests/util/test_config_reader.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Generator
+import pytest
+
+from instana.util.config import parse_ignored_endpoints_from_yaml
+
+
+class TestConfigReader:
+    @pytest.fixture(autouse=True)
+    def _resource(self) -> Generator[None, None, None]:
+        yield
```
pvital, line 11: You can remove this _resource() function if no pre-setting is necessary.

tests/util/test_config_reader.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def test_load_file(self, caplog: pytest.LogCaptureFixture) -> None:
```
pvital, line 13: I am not a fan of having a single test case with all possible cases written (or worse, duplicated) on it.
Please, create one test case using the parametrized feature of `pytest`, or two or more tests for each case.

tests/util/test_configuration-1.yaml
pvital, line 1: Missing copyright headers here.

tests/util/test_configuration-2.yaml
pvital, line 1: Missing copyright headers here.

pvital (changes requested):
Few requests.
I am missing some tests for the new functions or environment variables to be parsed, as the `INSTANA_IGNORE_ENDPOINTS_PATH`.

PR #719, opened 2025-04-22
fix: added protobuf constraint to requirements.txt

CagriYonca:
grpcio-tools==1.71.0 supports only protobuf<6.0, yet.


pvital (approved):
It looks good to me.

PR #720, opened 2025-04-28
Add tests to verify activating the tracer without code changes

GSVarsha:
### Description

Only checking if `instana` is present in `sys.modules` and not explicitly recording/checking if any spans are recorded because any import like `from instana.<module> import <method>` loads auto-instrumentation [here](https://github.com/instana/python-sensor/blob/main/src/instana/__init__.py#L255) and hence failing the purpose of the intended tests.


pvital (approved):
It looks good to me.

PR #725, opened 2025-05-02
fix: change request url for urllib

CagriYonca:
The old service (Reqres) started to require an API key on the 25th of April


GSVarsha (approved):
Looks good to me!

PR #727, opened 2025-05-07
fix: Function type has been updated

CagriYonca:
Updated the basic_consume wrapper function type to async


pvital (approved):
It looks good to me.

PR #729, opened 2025-05-07
fix: Prevent internal span creation when root exit variable is set

CagriYonca:
Internal spans are being created by internal http calls of the host. I've adapted urllib3 to prevent this issue.

### DIFF
--- a/src/instana/instrumentation/urllib3.py
+++ b/src/instana/instrumentation/urllib3.py
@@ -11,7 +11,11 @@ from instana.log import logger
 from instana.propagators.format import Format
 from instana.singletons import agent
 from instana.util.secrets import strip_secrets_from_query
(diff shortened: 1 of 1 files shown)


src/instana/instrumentation/urllib3.py
```diff
@@ -91,7 +95,11 @@ def urlopen_with_instana(
         tracer, parent_span, span_name = get_tracer_tuple()
 
         # If we're not tracing, just return; boto3 has it's own visibility
-        if tracing_is_off() or (span_name == "boto3"):
+        if (
+            tracing_is_off()
+            or (span_name == "boto3")
+            or (agent.options.allow_exit_as_root and "com.instana" in kwargs.get("url"))
```
pvital, line 101: That works, but it will block all spans pointing to any URL containing "com.instana", and this can cause missing spans in the future.

Why not revisit the PR #562 and supress all internal calls?

PR #738, opened 2025-05-21
fix: updated amqp instrumentation, adapted attributes and unittests

CagriYonca:
- Updated aioamqp instrumentation, changed spantype from sdk to amqp. 
- Adapted attributes according to the backend code.

### DIFF
--- a/src/instana/instrumentation/aioamqp.py
+++ b/src/instana/instrumentation/aioamqp.py
@@ -26,11 +26,29 @@ try:
             "aioamqp-publisher", span_context=parent_context
         ) as span:
             try:
-                span.set_attribute("aioamqp.exchange", argv[0])
-                return await wrapped(*argv, **kwargs)
+                span.set_attribute("amqp.command", "publish")
+                span.set_attribute("amqp.routing_key", kwargs.get("routing_key"))
+
+                protocol = getattr(instance, "protocol", None)
+                transport = getattr(protocol, "_transport", None)
+                extra = getattr(transport, "_extra", {}) if transport else {}
+                peername = extra.get("peername")
+                if (
+                    peername
+                    and isinstance(peername, (list, tuple))
+                    and len(peername) >= 2
+                ):
+                    connection_info = f"{peername[0]}:{peername[1]}"
+                else:
+                    connection_info = "unknown"
+                span.set_attribute("amqp.connection", connection_info)
+
+                response = await wrapped(*argv, **kwargs)
             except Exception as exc:
                 span.record_exception(exc)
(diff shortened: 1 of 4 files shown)


src/instana/instrumentation/aioamqp.py
```diff
@@ -26,11 +26,31 @@ async def basic_publish_with_instana(
...
+                    connection_info = f"{peername[0]}:{peername[1]}"
+                else:
+                    connection_info = "unknown"
+                span.set_attribute("amqp.connection", connection_info)
+
+                response = await wrapped(*argv, **kwargs)
             except Exception as exc:
                 span.record_exception(exc)
                 logger.debug(f"aioamqp basic_publish_with_instana error: {exc}")
+            finally:
```
pvital, line 52: You must use `else` here. 

The `finally` block executes if both the `try` and  `exception` blocks are executed. However, you don't set the `response` variable in the `exception` block, so it will raise an `UnboundLocalError`.

The `else` block executes only if the `try` block executes with no errors.

src/instana/instrumentation/aioamqp.py
```diff
@@ -58,14 +78,32 @@ async def callback_wrapper(
...
+                        connection_info = f"{peername[0]}:{peername[1]}"
+                    else:
+                        connection_info = "unknown"
+                    span.set_attribute("amqp.connection", connection_info)
+
+                    response = await wrapped_callback(*args, **kwargs)
                 except Exception as exc:
                     span.record_exception(exc)
                     logger.debug(f"aioamqp basic_consume_with_instana error: {exc}")
+                finally:
```
pvital, line 105: Same as before.

pvital (changes requested):
Few comments.

tests/frameworks/test_aioamqp.py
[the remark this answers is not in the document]
CagriYonca: Done

src/instana/instrumentation/aioamqp.py
```diff
@@ -26,11 +26,31 @@ async def basic_publish_with_instana(
             "aioamqp-publisher", span_context=parent_context
         ) as span:
             try:
-                span.set_attribute("aioamqp.exchange", argv[0])
-                return await wrapped(*argv, **kwargs)
+                span.set_attribute("amqp.message", argv[0])
```
GSVarsha, line 29: I couldn't find an entry for `amqp.message` attribute in the backend/ui-client code. Could you please confirm if we need this here? 
If yes, can you `decode()` the message from `bytes` to `str` and then store?

pvital (approved):
It looks good to me.

PR #745, opened 2025-06-10
currency: aiohttp version is bounded for python 3.8

CagriYonca:
Adding bounded limits to aiohttp package for Python 3.8


pvital (approved):
It looks good to me.

PR #746, opened 2025-06-16
fix(currency-report): Handle the edge case while matching the dependencies

GSVarsha:
- [report(currency): simplify release date extraction](https://github.com/instana/python-sensor/commit/c18a9c1839b98d17f97569e43b4c4143bb2247f3)
- [fix(currency-report): Handle edge case while matching the dependency](https://github.com/instana/python-sensor/commit/a30f2fdbebc5401fe719596dce035615131df8b0)
- [chore: Add spyne, aio-pika & aioamqp to currency report](https://github.com/instana/python-sensor/commit/7c51a90f7a004290f66aec43d3831ed9f4714435)


pvital (approved):
It looks good to me.

PR #749, opened 2025-06-18
Fix: CI errors with sanic

GSVarsha:
``` python
venv/lib/python3.9/site-packages/sanic/pages/error.py:4: in <module>
    import tracerite.html
venv/lib/python3.9/site-packages/tracerite/__init__.py:1: in <module>
    from .html import html_traceback
venv/lib/python3.9/site-packages/tracerite/html.py:3: in <module>
    from .trace import extract_chain
venv/lib/python3.9/site-packages/tracerite/trace.py:11: in <module>
    from .inspector import extract_variables
E     File "/root/repo/venv/lib/python3.9/site-packages/tracerite/inspector.py", line 119
E       except AttributeError, TypeError:
E                            ^
E   SyntaxError: invalid syntax
```

Error comes from [this](https://github.com/sanic-org/tracerite/pull/17/files#diff-5d409ed00bbdfa1af787bb228eada817ee27013f88370edde4698bc86f9ad9b0R122) change that was released in the latest version (1.1.2) 5 hours ago in `tracerite` (`sanic`'s dependency)

### DIFF
--- a/tests/requirements-pre314.txt
+++ b/tests/requirements-pre314.txt
@@ -31,10 +31,9 @@ pytz>=2024.1
 redis>=3.5.3
 requests-mock
 responses<=0.17.0
-# Sanic is not installable on 3.13 because `httptools, uvloop` dependencies fail to compile:
-# `too few arguments to function ‘_PyLong_AsByteArray’`
-sanic>=19.9.0; python_version < "3.13"
-sanic-testing>=24.6.0; python_version < "3.13"
+# Sanic doesn't support python-3.14 yet
+# sanic>=19.9.0
+# sanic-testing>=24.6.0
 starlette>=0.38.2
 sqlalchemy>=2.0.0
 tornado>=6.4.1
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -50,16 +50,13 @@ if sys.version_info >= (3, 12):
     collect_ignore_glob.append("*test_spyne*")
 
 
-if sys.version_info >= (3, 13):
-    # Currently not installable dependencies because of 3.13 incompatibilities
-    collect_ignore_glob.append("*test_sanic*")
-
-
 if sys.version_info >= (3, 14):
     # Currently not installable dependencies because of 3.14 incompatibilities
     collect_ignore_glob.append("*test_fastapi*")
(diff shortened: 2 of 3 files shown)


arjun-rajappa (commented):
Should we add this logic to the dependency of python-sensor package? If someone uses sanic and tries to instrument it we may get the same error!

tests/requirements-pre314.txt
```diff
@@ -31,10 +31,6 @@ pytz>=2024.1
 redis>=3.5.3
 requests-mock
 responses<=0.17.0
-# Sanic is not installable on 3.13 because `httptools, uvloop` dependencies fail to compile:
-# `too few arguments to function ‘_PyLong_AsByteArray’`
-sanic>=19.9.0; python_version < "3.13"
-sanic-testing>=24.6.0; python_version < "3.13"
```
pvital, line 37: I don't think removing those lines from the Python 3.14 requirements file is necessary, since we have an upper bound constraint.

However, since you are fixing something related to Sonic, it would be good to check if it is installing in 3.13 now. It's more than 6 months after the 3.13 release, and for sure, it has support at this moment.

pvital (changes requested):

pvital:
> Should we add this logic to the dependency of python-sensor package? If someone uses sanic and tries to instrument it we may get the same error!

I don't think it is necessary, since Sanic is not a direct dependency. 

It's a supported framework, and if this is coming from a third-party dependency, probably this will be handled by Sanic or the original package.

tests/requirements-pre314.txt
GSVarsha: > I don't think removing those lines from the Python 3.14 requirements file is necessary, since we have an upper bound constraint.

I've removed the upper bound constraint since this file is used only for `python-3.14` and commented the lines instead.

> However, since you are fixing something related to Sonic, it would be good to check if it is installing in 3.13 now.

Thanks for pointing this out! I've checked on both `3.13` and `3.14`, looks like the former is indeed supported and the latter is not. Confirmed the same in sanic's docs. Updating the PR accordingly :)

pvital (approved):
It looks good to me.

PR #752, opened 2025-06-27
feat: added single method option to configuration.yaml

CagriYonca:
Added similar functionality below
https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-configuration#ignoring-endpoints

### DIFF
--- a/src/instana/util/config.py
+++ b/src/instana/util/config.py
@@ -82,16 +82,24 @@ def parse_endpoints_of_service(
     """
     if service == "kafka" and isinstance(methods, list):
         for rule in methods:
-            for method, endpoint in itertools.product(
-                rule["methods"], rule["endpoints"]
-            ):
-                ignore_endpoints.append(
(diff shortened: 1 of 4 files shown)


src/instana/util/config.py
```diff
@@ -82,18 +82,26 @@ def parse_endpoints_of_service(
...
-                    f"{service.lower()}.{method.lower()}.{endpoint.lower()}"
-                )
+            ignore_endpoints.extend(parse_kafka_methods(rule))
     else:
         for method in methods:
             ignore_endpoints.append(f"{service.lower()}.{method.lower()}")
     return ignore_endpoints
 
 
+def parse_kafka_methods(rule: Union[str, Dict[str, any]]):
```
pvital, line 92: I loved it!!! You are making the code cleaner ;-) 
However, you forgot to add the type annotation for the return of the function.

pvital (changes requested):

pvital (approved):
It looks good to me.

PR #756, opened 2025-07-01
feat: feat: Add `httpx` instrumentation 

GSVarsha:
feat: feat: Add `httpx` instrumentation 
- Add instrumentation to `sync` clients
- Add instrumentation to `async` clients
- Add unit tests for both `sync` and `async` requests
- Adapt current unit tests to support `httpx`


pvital (approved):
It looks good to me.

PR #758, opened 2025-07-01
Add new AWS region to publish script.

pvital:
Added the new AWS region `ap-east-2` to publish script.


GSVarsha (approved):
Looks good to me

PR #759, opened 2025-07-02
Update InstanaTracer to utilize kind argument

HonakerM:
Hello,

I am working on instrumenting an application with instana and I was really confused on why the `kind` argument was not working for setting span kinds. I then realized it's because it's not used! This PR fixes that and automatically adds the kind if one was provided

### DIFF
--- a/src/instana/tracer.py
+++ b/src/instana/tracer.py
@@ -138,6 +138,10 @@ class InstanaTracer(Tracer):
             # events: Sequence[Event] = None,
         )
 
+        # Auto add span kind if provided
+        if kind:
+            span.set_attribute("span.kind", kind)
+
         if name in EXIT_SPANS:
             self._add_stack(span)
 


HonakerM:
CC: @GSVarsha @pvital @CagriYonca wasn't sure who I should ping for review/merge

pvital (changes requested):
Hello @HonakerM. Thank you for your PR.

We don't "use" the kind while starting the span, because we set this value following our internal table of types based on the span name during the closing stage of the span lifetime, when we create the Instana Span payload before transmiting it to the Agent. Check the `RegisteredSpan` class and you will see this setting.

The `start_span()` and the `start_as_current_span()` have the `kind` as argument because we are implementing the OpenTelemetry's Interface for the Tracer.

Your contribution is very useful, but still need a bit more work since it will be overwrited by the `RegisteredSpan` logic. In addition, changes in the unit tests will be necessary and more than welcomed.

PR #763, opened 2025-07-15
fix(aiohttp): remove support for older versions of aiohttp

CagriYonca:
https://github.com/instana/python-sensor/security/dependabot/33

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,10 +34,12 @@ collect_ignore_glob = [
 # ppc64le and s390x have limitations with some supported libraries.
 machine, py_version = get_runtime_env_info()
 if machine in ["ppc64le", "s390x"]:
-    collect_ignore_glob.extend([
-        "*test_google-cloud*",
-        "*test_pymongo*",
-    ])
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -51,28 +53,40 @@
...
 if not os.environ.get("KAFKA_TEST"):
     collect_ignore_glob.append("*kafka/test*")
 
+# Currently asyncio and tornado_server depends on aiohttp and
+# since aiohttp versions < 3.12.14 have vulnerability we skip the tests below
+if sys.version_info < (3, 9):
+    collect_ignore_glob.append("*test_aiohttp*")
+    collect_ignore_glob.append("*test_asyncio*")
+    collect_ignore_glob.append("*test_tornado_server*")
+
```
pvital, line 72: Please start using the `collect_ignore_glob.extend([])` as in the other blocks.

```suggestion
if sys.version_info < (3, 9):
    collect_ignore_glob.extend([
        "*test_aiohttp*",
        "*test_asyncio*",
        "*test_tornado_server*",
    ])

```

PR #764, opened 2025-07-17
Windows support.

pvital:
This PR adds support to Windows OS. It does the following:

- Upgrades the collection of runtime environment info, by changing the `util get_runtime_env_info()` function also to return the system/OS information about the current runtime environment.
- Creates three new functions to return if the current runtime environment runs on a Windows system, and ppc64 or s390x architectures.
- Adds Windows platform support for resource metrics by creating a new `resource_usage.py` module with platform-specific implementations.
- Adds `psutil` dependency for Windows systems in `pyproject.toml`.
- Refactor `runtime.py` to use the new cross-platform resource usage functions, and
- Implements the `ResourceUsage` class that provides consistent interface across platforms.

The implementation gracefully handles platform differences, ensuring consistent metrics collection on both Unix and Windows environments.


src/instana/collector/helpers/resource_usage.py
```diff
@@ -0,0 +1,126 @@
...
+
+import os
+from typing import NamedTuple
+
+from instana.log import logger
+from instana.util.runtime import is_windows
+
+
+class ResourceUsage(NamedTuple):
+    """Cross-platform resource usage information"""
```
GSVarsha, line 13: Can we add a docstring that explains what each of these attributes are?
    
    """
    Cross-platform resource usage information, mirroring fields found in the Unix rusage struct.

    Attributes:
        ru_utime (float): User CPU time used (seconds).
        ru_stime (float): System CPU time used (seconds).
        ru_maxrss (int): Maximum resident set size used (bytes).
        ru_ixrss (int): Integral shared memory size (bytes).
        ru_idrss (int): Integral unshared data size (bytes).
        ru_isrss (int): Integral unshared stack size (bytes).
        ru_minflt (int): Number of page reclaims (soft page faults).
        ru_majflt (int): Number of page faults requiring I/O (hard page faults).
        ru_nswap (int): Number of times a process was swapped out.
        ru_inblock (int): Number of file system input blocks.
        ru_oublock (int): Number of file system output blocks.
        ru_msgsnd (int): Number of messages sent.
        ru_msgrcv (int): Number of messages received.
        ru_nsignals (int): Number of signals received.
        ru_nvcsw (int): Number of voluntary context switches.
        ru_nivcsw (int): Number of involuntary context switches.
    """

GSVarsha (approved):
Just a small docstring suggestion. 
Looks perfect to me otherwise!

src/instana/collector/helpers/resource_usage.py
pvital: Done.
I added you as a co-author of the commit for this suggestion ;-)

PR #766, opened 2025-07-24
Remove six package and its usage since Python 2 is no longer supported

GSVarsha:
The package `six` was used to bridge the gap between Python 2 and Python 3 but since we have removed support for Python 2 long back, it would only make sense to completely remove its usage from our codebase.

### DIFF
--- a/tests/clients/test_google-cloud-storage.py
+++ b/tests/clients/test_google-cloud-storage.py
@@ -14,7 +14,7 @@ from tests.test_utils import _TraceContextMixin
 from opentelemetry.trace import SpanKind
 
 from mock import patch, Mock
-from six.moves import http_client
+from http import client as http_client
 
 from google.cloud import storage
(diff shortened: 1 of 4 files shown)


tests/clients/test_google-cloud-storage.py
```diff
@@ -14,7 +14,7 @@
 from opentelemetry.trace import SpanKind
 
 from mock import patch, Mock
-from six.moves import http_client
+import http.client as http_client
```
pvital, line 17: Following your Slack poll result:

```suggestion
from http import client as http_client
```

pvital (changes requested):
One minor request.

tests/clients/test_google-cloud-storage.py
GSVarsha: My bad, forgot to commit this change after the poll.

PR #768, opened 2025-07-29
Disabling log collection at the tracer level

pvital:
We need to disable log collection at the tracer level to prevent duplication in the backend when both the application tracer and an OpenTelemetry collector are running on the same system.

This PR adds support for the Span Disabling feature, which allows the exclusion of specific traces or calls from tracing based on the category (technology) or type (frameworks, libraries, instrumentations) supported by the traces.

In addition, it refactors the initialization of the Instana module and the ConfigReader class.

### DIFF
--- a/src/instana/__init__.py
+++ b/src/instana/__init__.py
@@ -20,6 +20,7 @@ from instana.collector.helpers.runtime import (
     is_autowrapt_instrumented,
     is_webhook_instrumented,
 )
+from instana.util.config import is_truthy
 from instana.version import VERSION
 
 __author__ = "Instana Inc."
@@ -225,7 +226,9 @@ def _start_profiler() -> None:
         profiler.start()
 
 
-if "INSTANA_DISABLE" not in os.environ:
+if "INSTANA_DISABLE" not in os.environ and not is_truthy(
+    os.environ.get("INSTANA_TRACING_DISABLE", None)
+):
     # There are cases when sys.argv may not be defined at load time.  Seems to happen in embedded Python,
     # and some Pipenv installs.  If this is the case, it's best effort.
     if (
--- a/src/instana/__init__.py
+++ b/src/instana/__init__.py
@@ -11,6 +11,7 @@ Source Code: https://github.com/instana/python-sensor
 """
 
 import importlib
+import importlib.util
 import os
 import sys
 from typing import Tuple
@@ -70,7 +71,7 @@ def load(_: object) -> None:
 def apply_gevent_monkey_patch() -> None:
     from gevent import monkey
 
-    if os.environ.get("INSTANA_GEVENT_MONKEY_OPTIONS"):
+    if provided_options := os.environ.get("INSTANA_GEVENT_MONKEY_OPTIONS"):
 
         def short_key(k: str) -> str:
             return k[3:] if k.startswith("no-") else k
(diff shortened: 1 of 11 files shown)


tests/util/test_configuration-2.yaml
```diff
@@ -18,3 +18,7 @@ com.instana.tracing:
         endpoints: ["span-topic", "topic4"]
       # - methods: ["consume", "send"]
       #   endpoints: ["*"] # Applied to all topics
+  disable:
+    - "logging": true
+    - "databases": true
+    - "redis": false
```
GSVarsha, line 24: Same as https://github.com/instana/python-sensor/pull/768/files#r2241760506

GSVarsha (changes requested):
Just a few requests. Looks good to me otherwise.

tests/clients/test_logging.py
```diff
@@ -168,3 +168,78 @@ def main():
         assert spans[0].k is SpanKind.CLIENT
 
         assert spans[0].data["log"].get("message") == "foo bar"
+
+
+class TestLoggingDisabling:
```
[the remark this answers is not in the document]
pvital: Yeah, Bob is doing a fantastic job. 😒 Me either, especially working so late. 
I will fix those and let you know when it's done. Thanks for noticing them.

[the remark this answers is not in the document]
pvital: Done.

tests/util/test_configuration-1.yaml
```diff
@@ -17,3 +17,7 @@ tracing:
         endpoints: ["span-topic", "topic4"]
       # - methods: ["consume", "send"]
       #   endpoints: ["*"] # Applied to all topics
+  disable:
+    - "logging": true
+    - "databases": true
+    - "redis": false
```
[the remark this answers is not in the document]
pvital: Done

src/instana/__init__.py
```diff
@@ -218,14 +209,26 @@ def boot_agent() -> None:
...
-if "INSTANA_DISABLE" not in os.environ:
+def _start_profiler() -> None:
+    """Start the Instana Auto Profile."""
+    from instana.singletons import get_profiler
+
+    if profiler := get_profiler():
+        profiler.start()
+
+
+if "INSTANA_DISABLE" not in os.environ and not is_truthy(
```
GSVarsha, line 229: Just realised that we haven't documented any environmental variable configuration named `INSTANA_DISABLE`.
This also makes me wonder what the difference is between the following.

`export INSTANA_DISABLE_AUTO_INSTR="true"`
   vs
`export INSTANA_TRACING_DISABLE="true"`

src/instana/__init__.py
```diff
@@ -11,6 +11,7 @@
 """
 
 import importlib
+import importlib.util
```
GSVarsha, line 14: Not sure why even bob recommends `import importlib.util` over `from importlib import util` 😅

src/instana/__init__.py
pvital: There is no difference between `INSTANA_DISABLE` and `INSTANA_TRACING_DISABLE`: both will completely disable the creation of spans, actually, the load of the rest of our library. `INSTANA_DISABLE` continues there for legacy reasons, I guess.

Searching the Instana documentation, I did not find any reference to this environment variable, so I will remove it from our code. What do you think, @GSVarsha? 

Now, the `INSTANA_DISABLE_AUTO_INSTR` is documented [here for Python](https://www.ibm.com/docs/en/instana-observability/1.0.301?topic=python-configuration-configuring-instana-package#disabling-automatic-instrumentation) and [here](https://www.ibm.com/docs/en/instana-observability/1.0.301?topic=references-environment-variables#environment-variables-for-dynamic-language-sensors) as a general environment variable.  It disables automatic instrumentation - used in K8s environments where you don't want the instrumentation from the Autotrace webhook.

src/instana/__init__.py
pvital: Probably because the code using it was `importlib.util.find_spec("gevent")`.

I changed to `from importlib import util as importlib_util` and its usage to `importlib_util.find_spec("gevent")` to differentiate it from other `util` imports.

GSVarsha (approved):
Looks good to me!

Issue #769, opened 2025-07-30
[Bug]: Instana silently breaks s3 upload_fileobj Api


HonakerM:
### Problem Description

Hello, when using instana with the `boto3.s3.Bucket.upload_fileobj` api we start silently failing to upload objects.

The following code works on it's own if an s3 server is listening on `localhost:9000`

s3 setup using minio if needed:
```shell
 docker run -p 9000:9000 -p 9001:9001  --name minio -d minio/minio server /data --console-address ":9001"
```

Working python code:
```python
from io import BytesIO

import boto3
s3_resource = boto3.resource(
    service_name="s3",
    region_name=None,
    use_ssl=True,
    verify=True,
    endpoint_url="http://127.0.0.1:9000",
    aws_access_key_id="minioadmin",
    aws_secret_access_key="minioadmin",
)
bucket = s3_resource.Bucket(name="test")
bucket.upload_fileobj(BytesIO(b"somedata"), "somekey")
response = bucket.Object("somekey").get()
file_content = response["Body"].read()
assert file_content == b"somedata"
```

If I change this code to have an instana span like the following. I start to get an assert error:
```python
from io import BytesIO
from instana.singletons import tracer
import boto3
s3_resource = boto3.resource(
    service_name="s3",
    region_name=None,
    use_ssl=True,
    verify=True,
    endpoint_url="http://127.0.0.1:9000",
    aws_access_key_id="minioadmin",
    aws_secret_access_key="minioadmin",
)
bucket = s3_resource.Bucket(name="test")

with tracer.start_as_current_span("test"):
    bucket.upload_fileobj(BytesIO(b"somedata"), "somekey")
response = bucket.Object("somekey").get()
file_content = response["Body"].read()
assert file_content == b"somedata"
```

There are two parts that are concerning to me:

1. The error in `upload_fileobj` is completely hidden. The function returns like normal but the upload never happens
2. There is an error at all. Instana traces should not affect the underlying package operation

### Python Version

Python 3.11

### Python Modules

```shell
Due to security I can't list all packages. Here is the affected versions used in the above example:


instana==3.4.1
boto3==1.35.78
botocore==1.35.78
```

### Python Environment

```shell
Unable due to security. None of my envs start with `INSTANA` or have any affect on the above example
```

GSVarsha:
Hi @HonakerM , thanks for reaching out! 
I will take a look at this bug and your fix PR, and let you know if any other changes or additions are needed.

PR #770, opened 2025-07-30
Update AWS wrapper to not hide proper exceptions

HonakerM:
This PR fixes the errors described in https://github.com/instana/python-sensor/issues/769.

Not only does it update the code to properly fetch the bucket name for upload_fileobj it also ensures we don't hide proper exceptions from the client


GSVarsha (commented):
Hi @HonakerM, we would require relevant testcases before merging this change. If you don't already have one, I can add them to this PR.

PR #774, opened 2025-08-06
Remove Python 3.8 from the project dependencies

GSVarsha:
Remove Python 3.8 from the project dependencies
- install `typing_extensions` only for Python < 3.11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -304,7 +304,7 @@ workflows:
       - python3x:
           matrix:
             parameters:
-              py-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+              py-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
       - python314
       - py39cassandra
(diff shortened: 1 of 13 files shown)


src/instana/version.py
```diff
@@ -3,4 +3,4 @@
 
 # Module version file.  Used by setup.py and snapshot reporting.
 
-VERSION = "3.5.0"
+VERSION = "3.5.0.post1"
```
pvital, line 6: Shouldn't be `3.5.1-fedramp` ?

```suggestion
VERSION = "3.5.1-fedramp"
```

pvital (changes requested):
Small question.

src/instana/version.py
GSVarsha: Hi @pvital , related slack conversation: [link](https://instana.slack.com/archives/C06A6H9EKHS/p1754473681812429?thread_ts=1754473476.101859&cid=C06A6H9EKHS)

PR #782, opened 2025-08-11
fix: stop notifying draft PRs

GSVarsha:


.github/workflows/opened-pr-notification-on-slack.yml
```diff
@@ -6,7 +6,7 @@ permissions:
 
 on:
   pull_request:
-    types: [opened, reopened]
+    types: [reopened, review_requested]
```
pvital, line 9: If `reopened` is used with a list of reviewers, it executes the workloaod twice - once for each event. 

I guess it would be interesting to keep only the `opened` and `reopened` and then check the value of the `github.event.pull_request.draft` as listed [here](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=opened#pull_request).

Something like:

```suggestion
    types: [opened, reopened]
 
jobs:
  notify-slack:
    runs-on: ubuntu-latest
    if: ${{ !github.event.pull_request.draft }}

```

pvital (changes requested):

.github/workflows/opened-pr-notification-on-slack.yml
GSVarsha: Thanks for the suggestion, Paulo! 

I will add `ready_for_review` as well because currently if the pull request is later marked "Ready for review" from a draft state, it doesn't trigger the `opened` or `reopened` events. 

```suggestion
    types: [opened, reopened, ready_for_review]
 
jobs:
  notify-slack:
    runs-on: ubuntu-latest
    if: ${{ !github.event.pull_request.draft }}
```

PR #786, opened 2025-08-19
ci: Update Tekton pipelines

pvital:
1. Updated Python versions and container images tags.
2. Updated PR pipeline to test all necessary cases.
3. Updated scheduled pipelines. Now the currency pipeline executes after 1h30 the start of the daily pipeline.


CagriYonca (approved):
Looks good to me

PR #789, opened 2025-08-23
Internal Span Filtering

CagriYonca:
Add internal span filtering for internal HTTP calls.


pvital (approved):
It looks good to me.  

As we discussed internally, let's modify this in the near future to use the new `Span Filtering` feature.

PR #790, opened 2025-08-23
Kafka FUP Final

CagriYonca:
Fixed trace correlation issues, aligned with other tracers through the tracer-test-suite.

### DIFF
--- a/src/instana/instrumentation/kafka/confluent_kafka_python.py
+++ b/src/instana/instrumentation/kafka/confluent_kafka_python.py
@@ -1,19 +1,27 @@
 # (c) Copyright IBM Corp. 2025
 
+
 try:
+    import contextvars
     from typing import Any, Callable, Dict, List, Optional, Tuple
 
     import confluent_kafka  # noqa: F401
     import wrapt
     from confluent_kafka import Consumer, Producer
+    from opentelemetry import context, trace
     from opentelemetry.trace import SpanKind
 
     from instana.log import logger
     from instana.propagators.format import Format
+    from instana.singletons import get_tracer
     from instana.util.traceutils import (
         get_tracer_tuple,
         tracing_is_off,
     )
+    from instana.span.span import InstanaSpan
+
+    consumer_token = None
+    consumer_span = contextvars.ContextVar("confluent_kafka_consumer_span")
 
     # As confluent_kafka is a wrapper around the C-developed librdkafka
     # (provided automatically via binary wheels), we have to create new classes
@@ -46,6 +54,9 @@ try:
             self, timeout: Optional[float] = -1
         ) -> Optional[confluent_kafka.Message]:
             return super().poll(timeout)
+
+        def close(self) -> None:
+            return super().close()
 
     def trace_kafka_produce(
         wrapped: Callable[..., InstanaConfluentKafkaProducer.produce],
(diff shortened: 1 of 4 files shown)


CagriYonca:
A couple of words behind the need for the changes.

Let's say we have two distributed environments: `env A` as a producer and `env B` as a consumer, and say we have a flask endpoint named send_message(), which receives the call and sends a message to a kafka topic, after sending the message we also make an http call to let's say `ibm-producer-api`. the consumer checks the topic every 10 seconds, reads the message if any, then makes another http call to `ibm-consumer-api`. 

So the pseudo code is here, trace IDs are given as examples:

env A - producer
- send_message() endpoint receives a call, WSGI Span is created, trace ID: 100
  - message is sent to kafka topic, kafka-produce span is created, trace ID: 100
  - http call made to `ibm.com`, urllib3 span is created, trace ID: 100

env B - consumer
- consumer checks the topic every 10 seconds for any new messages, if any:
  - message is consumed from the kafka topic, kafka-consume span is created, trace ID: 100
  - since there is no active span after the message has been consumed, kafka-consume span is ended, and since the span context was not saved, we lose the trace ID, so the new urllib3 span will have the trace ID: 101

What we want is whether there is a parent span or not, to keep the incoming trace ID alive. 

So after this PR, what will happen in the example above is as follows:

env B - consumer
- consumer checks the topic every 10 seconds for any new messages, if any:
  - message is consumed from the kafka topic, kafka-consume span is created, trace ID: 100
  - kafka-consume span is still alive until we call `kafka_consumer.close()` or manually close the span using `close_consumer_span()`. now since the span context information is still there, urllib3 span will also have the trace ID: 100

src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -105,78 +113,142 @@ def create_span(
...
-        finally:
-            if res:
-                for message in res:
-                    create_span("consume", message.topic(), message.headers())
-            else:
-                create_span("consume", exception=exception)
+            create_span("consume", exception=exception)
+
+    def trace_kafka_close(
+        wrapped: Callable[..., InstanaConfluentKafkaConsumer.close],
```
pvital, line 207: You have to add this function to the class `InstanaConfluentKafkaConsumer` to wrap it correctly.

src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -105,78 +113,142 @@ def create_span(
...
+        kwargs: Dict[str, Any],
+    ) -> None:
+        try:
+            # Close any existing consumer span before closing the consumer
+            span = consumer_span.get(None)
+            if span is not None:
+                close_consumer_span(span)
+
+            # Execute the actual close operation
+            result = wrapped(*args, **kwargs)
```
pvital, line 219: Use the variable name `res` to keep compatibility with the rest of the code.
```suggestion
            res = wrapped(*args, **kwargs)
```

src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -1,19 +1,27 @@
...
     from instana.log import logger
     from instana.propagators.format import Format
+    from instana.singletons import get_tracer
     from instana.util.traceutils import (
         get_tracer_tuple,
         tracing_is_off,
     )
+    from instana.span.span import InstanaSpan
+
+    consumer_token: Dict[str, Any] = {}
```
pvital, line 23: Is this variable necessary to be a dictionary?  I saw that only one value to the unique key `token` is assigned to it, so I think a simple variable is sufficient, don't you think?

BTW, we are not using type annotations for variables in our code. 

```suggestion
    consumer_token = None
```

src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -105,78 +113,142 @@ def create_span(
...
+                span.end()
+
+            save_consumer_span_into_context(span)
+        except Exception as e:
+            logger.debug(f"Error while creating kafka-consumer span: {e}")
+
+    def save_consumer_span_into_context(span: "InstanaSpan") -> None:
+        ctx = trace.set_span_in_context(span)
+        token = context.attach(ctx)
+        consumer_token["token"] = token
```
pvital, line 173: If changing the `consumer_token` variable as commented [here](https://github.com/instana/python-sensor/pull/790/files#r2302190833), you can do the following:
```suggestion
        consumer_token = context.attach(ctx)
```

src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -105,78 +113,142 @@ def create_span(
...
+        consumer_span.set(span)
+
+    def close_consumer_span(span: "InstanaSpan") -> None:
+        if span.is_recording():
+            span.end()
+            consumer_span.set(None)
+        if "token" in consumer_token:
+            context.detach(consumer_token.pop("token", None))
```
pvital, line 181: If changing the `consumer_token` variable as commented [here](https://github.com/instana/python-sensor/pull/790/files#r2302190833), you can do the following:
```suggestion
        if consumer_token:
            context.detach(consumer_token)
```

src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -105,78 +113,142 @@ def create_span(
...
+
+    def clear_context() -> None:
+        context.attach(trace.set_span_in_context(None))
+        consumer_token.clear()
```
pvital, line 185: If changing the `consumer_token` variable as commented [here](https://github.com/instana/python-sensor/pull/790/files#r2302190833), you can do the following:
```suggestion
        consumer_token = None
```

src/instana/instrumentation/kafka/kafka_python.py
```diff
@@ -1,23 +1,31 @@
...
     from instana.util.traceutils import (
         get_tracer_tuple,
         tracing_is_off,
     )
+    from instana.span.span import InstanaSpan
 
     if TYPE_CHECKING:
         from kafka.producer.future import FutureRecordMetadata
 
+    consumer_token: Dict[str, Any] = {}
```
pvital, line 26: The same comment as [here](https://github.com/instana/python-sensor/pull/790/files#r2302190833).

BTW, there was too much repetitive code in those two files. It would be interesting to see how to reduce that.

src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -105,78 +113,142 @@ def create_span(
...
+            logger.debug("Kafka consumer closed and spans cleaned up")
+            return result
+
+        except Exception as exc:
+            logger.debug(f"Error while closing Kafka consumer: {exc}")
+            # Still try to clean up the span even if close fails
+            span = consumer_span.get(None)
+            if span is not None:
+                close_consumer_span(span)
+            raise
```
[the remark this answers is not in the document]
CagriYonca: You're right, I was gonna remove the raises from the code

PR #791, opened 2025-08-28
feat: Add GEvent instrumentation support with OTel

GSVarsha:
feat: Add GEvent instrumentation support with OTel


.circleci/config.yml
```diff
@@ -208,7 +208,7 @@ jobs:
       - store-pytest-results
       - store-coverage-report
 
-  py39gevent_starlette:
+  py39starlette:
```
pvital, line 211: You can remove this since Starlette is being tested in all Python3.x tests.  
See [here](https://app.circleci.com/pipelines/github/instana/python-sensor/4364/workflows/bd00e1a9-a7f3-46f1-85f8-ac6af901828b/jobs/36333?invite=true#step-105-60397_81) the output for Python 3.13.

pvital (changes requested):
One minor request.

PR #796, opened 2025-09-08
fix: fixed reading suppression header from message's headers

CagriYonca:
Added the missing suppression header reading part to the send message function.


GSVarsha (approved):
Looks good to me

PR #798, opened 2025-09-09
fix: suppression propagation in kafka

GSVarsha:
### Why?

> Suppression `level` made in the child context or in the parent context has priority over any non-suppressed `level` setting

But the updated `span_context.level` was never propagated to the `KAFKA_HEADER_KEY_L_S` header, hence the change.


pvital (changes requested):
simple decoration request.

PR #800, opened 2025-09-12
fix(deps): pin setuptools to <81.0.0 to avoid pkg_resources deprecation warning

pvital:
Fix the following error to prevent security scan analysis reports:

```
/opt/instana/instrumentation/python/3.9.0/wrapt/importer.py:94: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ commands:
             CHANGED_FILES=$(git diff --name-only origin/main...HEAD)
             
             # Check if any relevant files changed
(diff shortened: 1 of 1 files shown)


pvital:
@GSVarsha, thanks for the review, but I have a quick question about this topic. Does your `wrapt` [PR](https://github.com/GrahamDumpleton/wrapt/pull/292) and related [discussion](https://github.com/GrahamDumpleton/wrapt/issues/291) fix this issue?

PR #801, opened 2025-09-22
(chore): AWS Lambda layer publishing script updates.

pvital:
This PR:

- Updates the compatible runtimes and arch for AWS Lambda layer publishing script.
- Updates dev region for AWS Lambda layer publishing script.
- Prints the AWS Lambda layer list as MarkDown.


GSVarsha (approved):
LGTM!

PR #808, opened 2025-10-07
Add support to Python 3.14.0

pvital:
This PR:

- Adds support for the new Python 3.14 version.
- Drops support for the Python 3.8 version.
- Updates CI pipelines to use the new 3.14 version and the containers based on the Debian `trixie` version instead of the `bookworm`.
- Since we have a 0-day policy for runtime support, it bumps up to version 3.9.0.


GSVarsha (approved):
Looks good to me

PR #815, opened 2025-10-28
Remove `setuptools` from project dependency

GSVarsha:
Remove `setuptools` from project dependency

### Why?
Dependency on `pkg_resources` gone after `wrapt-2.0.0`


pvital (approved):
It looks good to me.

PR #816, opened 2025-10-28
Fix Tuple IndexError issue in Kafka instrumentations.

pvital:
This PR fixes a Tuple IndexError in both `confluent_kafka_python.py` and `kafka_python.py` implementations.

It handles both positional and keyword arguments for the `topic` parameter in the respective `trace_kafka_produce` and `trace_kafka_send` functions. The issue occurred when the `topic` was passed as a keyword argument, resulting in an empty `args` tuple and causing an IndexError when trying to access `args[0]`.

1. **confluent_kafka_python.py**:
   - Fixed the `trace_kafka_produce` function to handle both positional and keyword arguments for the topic parameter
   - Added safety checks to handle edge cases
   - Added tests to verify the fix works with different argument patterns
   - Committed the changes with a detailed commit message

2. **kafka_python.py**:
   - Identified the same issue in the `trace_kafka_send` function
   - Implemented a similar fix to handle both positional and keyword arguments
   - Added tests to verify the fix works with different argument patterns
   - Committed the changes with a detailed commit message

Both fixes ensure that the Kafka instrumentation works correctly regardless of how the methods are called, improving the robustness of the Python sensor. The changes are minimal and focused on the specific issue, maintaining compatibility with existing code.

These changes should prevent the IndexError from occurring in the future, even when users pass the topic as a keyword argument instead of a positional argument.

### DIFF
--- a/src/instana/instrumentation/kafka/kafka_python.py
+++ b/src/instana/instrumentation/kafka/kafka_python.py
@@ -14,11 +14,8 @@ try:
     from instana.log import logger
     from instana.propagators.format import Format
     from instana.singletons import get_tracer
-    from instana.util.traceutils import (
-        get_tracer_tuple,
-        tracing_is_off,
-    )
     from instana.span.span import InstanaSpan
+    from instana.util.traceutils import get_tracer_tuple, tracing_is_off
 
     if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


GSVarsha (approved):
LGTM

PR #818, opened 2025-11-05
[INSTA-47091] fix: Improved getting active tracer and current span mechanism

CagriYonca:
`tracing_is_off` function is also widely used in almost every instrumentation and it only returns whether the tracer is ready to record a new span or not. As I already explained above, we already have this information in the tracer, so that by combining these two functions into get_tracer_tuple function, we're getting rid of 1 get_current_span call and functionality of `tracing_is_off` function since we can directly check if the tracer in the instrumentation is False or not.

- `get_active_tracer` and `tracing_is_off` functions has been removed
- adapted instrumentations to the new get_tracer_tuple function

### DIFF
--- a/src/instana/instrumentation/aio_pika.py
+++ b/src/instana/instrumentation/aio_pika.py
@@ -15,7 +15,7 @@ try:
 
     from instana.log import logger
     from instana.propagators.format import Format
-    from instana.util.traceutils import get_tracer_tuple, tracing_is_off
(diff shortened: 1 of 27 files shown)


GSVarsha (changes requested):
We shouldn't return if the tracer is inactive for server spans like wsgi, sanic, etc because those would (mostly) be the first spans (entry) starting the trace.

GSVarsha (approved):
Looks good to me!

PR #821, opened 2025-11-13
fix(wsgi): Ensure span stays active throughout the response iteration

GSVarsha:
### Long story short
Our existing WSGI middleware implementation did not consider the fact that WSGI applications return an **iterable** object.

### The long story
<html><head></head><body><span data-di="0" style="font-size:11pt"><div><span>A WSGI application must return an iterable of byte strings as its response body. </span><span>This iterable can be a list, a generator, or any other object that implements the iteration protocol. </span></div><br style="line-height:11pt"><div style="font-size:13pt;font-weight:bold;margin-bottom:11pt"><span>Here's why and how: </span></div><ul><li><span><span style="font-weight:bold">Streaming and Efficiency:</span> </span><span>Returning an iterable allows the WSGI server to send the response body in chunks, rather than waiting for the entire response to be generated in memory. </span><span>This is crucial for handling large responses, streaming data, and improving perceived performance for the client. </span></li><li><span><span style="font-weight:bold">Lazy Generation:</span> </span><span>By using a generator, the application can generate parts of the response body on demand, which is beneficial for resource-intensive operations or when dealing with dynamic content that changes over time. </span></li></ul><div style="font-size:13pt;font-weight:bold;margin-bottom:11pt">

Examples: 

• Returning a list: This is a common approach for smaller, complete responses. 

    def application(environ, start_response):
        status = '200 OK'
        headers = [('Content-type', 'text/plain')]
        start_response(status, headers)
        return [b"Hello, World!\n"]

• Using a generator: This allows for streaming and lazy generation. 

    def application(environ, start_response):
        status = '200 OK'
        headers = [('Content-type', 'text/plain')]
        start_response(status, headers)
        yield b"Hello "
        yield b"World!\n"

Important Considerations: 

• close() Method: If the returned iterable has a close() method, the WSGI server is responsible for calling it upon completion of the request, allowing the application to release any resources. [[1](https://grahamdumpleton.me/posts/2012/10/obligations-for-calling-close-on/)]

### DIFF
--- a/src/instana/version.py
+++ b/src/instana/version.py
@@ -3,4 +3,4 @@
 
 # Module version file.  Used by setup.py and snapshot reporting.
 
-VERSION = "3.9.2"
+VERSION = "3.9.3"


pvital (approved):
It looks good for me, even with space for a code cleanup. ;-)

PR #822, opened 2025-11-17
[INSTA-64769] chore: use get_tracer() for all tracer access

CagriYonca:
We should use the get_tracer function to provide the correct use case of the singleton pattern, so we can start by adapting our code to getter and setters. Using singletons with getters and setters may also be helpful in resolving span_queue access management issues that arise from multi-threading with Instana.

### DIFF
--- a/src/instana/instrumentation/flask/common.py
+++ b/src/instana/instrumentation/flask/common.py
@@ -1,25 +1,31 @@
-# (c) Copyright IBM Corp. 2021
+# (c) Copyright IBM Corp. 2021, 2025
 # (c) Copyright Instana Inc. 2019
 
 
-import wrapt
-import flask
+import re
 from importlib.metadata import version
-from typing import Callable, Tuple, Dict, Any, TYPE_CHECKING, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, Tuple, Type, Union
 
+import flask
+import wrapt
+from opentelemetry import context, trace
 from opentelemetry.semconv.trace import SpanAttributes
 
 from instana.log import logger
-from instana.singletons import tracer
 from instana.propagators.format import Format
-
+from instana.singletons import agent, get_tracer
+from instana.util.secrets import strip_secrets_from_query
+from instana.util.traceutils import extract_custom_headers
 
 if TYPE_CHECKING:
-    from werkzeug.exceptions import HTTPException
     from flask.typing import ResponseReturnValue
     from jinja2.environment import Template
+    from werkzeug.exceptions import HTTPException
 
-@wrapt.patch_function_wrapper('flask', 'templating._render')
+path_tpl_re = re.compile("<.*>")
+
+
+@wrapt.patch_function_wrapper("flask", "templating._render")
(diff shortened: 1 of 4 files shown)


src/instana/instrumentation/tornado/client.py
```diff
@@ -6,25 +6,25 @@
 
     import wrapt
     import functools
-    from typing import TYPE_CHECKING, Dict, Any
```
pvital: Just noticed this file is not "typed". 
Could you return this line and provide the type annotations we do? If you want, you can do it in a new commit.

src/instana/instrumentation/aio_pika.py
```diff
@@ -1,7 +1,6 @@
 # (c) Copyright IBM Corp. 2025
 
 try:
-    import aio_pika
```
pvital: Please do not remove the library imports; otherwise, we will not detect the packages we support during runtime. Instead, use the `# noqa: F401` pragma.

```python
    import aio_pika  # noqa: F401
```

tests/clients/boto3/test_boto3_lambda.py
```diff
@@ -7,7 +7,7 @@
 import boto3
 from moto import mock_aws
 
-from instana.singletons import tracer, agent
+from instana.singletons import get_tracer, agent
```
pvital:
```suggestion
from instana.singletons import agent, get_tracer
```

tests/frameworks/test_fastapi.py
```diff
@@ -5,7 +5,7 @@
 
 from fastapi.testclient import TestClient
 import pytest
-from instana.singletons import tracer, agent
+from instana.singletons import get_tracer, agent
```
pvital:
```suggestion
from instana.singletons import agent, get_tracer
```

example/carry_context.py
```diff
@@ -24,24 +27,24 @@ async def launch_async_calls(parent_span):
 
     # Now that we are inside of the event loop, first thing to do is to initialize
     # the tracing context using <parent_span> _and_ the asynchronous tracer <async_tracer>
-    with async_tracer.start_active_span('launch_async_calls', child_of=parent_span):
+    with async_tracer.start_active_span("launch_async_calls", child_of=parent_span):  # noqa: F821
```
GSVarsha: `async_tracer` is not defined here. I guess this file is outdated, we might have to migrate this to OTel specific examples in a different PR maybe.

GSVarsha (changes requested):
The Year 1 is the year IBM copyrighted the file content. If there's a new functionality added in the file (not bug fixes or structural changes), you can add the respective year if you want like the below:

```suggestion
# (c) Copyright IBM Corp. 2021, 2025
```

But I guess here we are only moving things around so I would suggest not to change.

example/carry_context.py
pvital: You can discard these changes in the `example` folder.
As I mentioned some time ago in a team meeting, I was thinking about deleting this folder from our repo.

PR #827, opened 2025-12-15
[Part 2] feat: Implement Span stack trace specification

GSVarsha:
### What's there in this PR?

Added stack-trace support for all configuration sources in the following precedence:

```
 ENV vars > YAML config (INSTANA_CONFIG_PATH) > in-code config > Agent config
```

### Runtime Stack Trace Collection Flow:

temp_image_1765877137141

> [!NOTE]
>  We support technology-specific overrides (for all `EXIT` spans!) for all the mapping based config sources (YAML, in-code and Agent)

### DIFF
--- a/src/instana/util/config.py
+++ b/src/instana/util/config.py
@@ -2,7 +2,7 @@
 
 import itertools
 import os
-from typing import Any, Dict, List, Sequence, Tuple, Union
+from typing import Any, Dict, List, Sequence, Tuple, Union, Optional
 
 from instana.configurator import config
 from instana.log import logger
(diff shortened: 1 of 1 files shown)


src/instana/util/config.py
```diff
@@ -295,21 +300,35 @@ def get_disable_trace_configurations_from_env() -> Tuple[List[str], List[str]]:
     return [], []
 
 
+def get_tracing_root_key(config_data: Dict[str, Any]) -> Union[str, None]:
```
pvital: It's a [best practice](https://docs.python.org/3/library/typing.html#typing.Optional) to use `Optional[X]` instead of `Union[X, None]`: 

```suggestion
def get_tracing_root_key(config_data: Dict[str, Any]) -> Optional[str]:
```

src/instana/util/config.py
```diff
@@ -319,4 +338,174 @@ def get_disable_trace_configurations_from_local() -> Tuple[List[str], List[str]]
     return [], []
 
 
+def validate_stack_trace_level(level_value: Any, context: str = "") -> Union[str, None]:
```
pvital: Same as [here](https://github.com/instana/python-sensor/pull/827/changes#r2626699246).

PR #832, opened 2026-01-07
Fysom improvement for multi-threaded operations

CagriYonca:
- The state machine was not compatible with multi-threads by default. Every fork was changing the state of the fysom so that Fysom was falling into `fysom.FysomError: event announce inappropriate in current state announced` error. After adding a thread-safe lock mechanism, it works and propagates spans as expected.
- Verified that the instrumentation works even without hooks.


src/instana/collector/base.py
```diff
@@ -70,9 +70,12 @@ def is_reporting_thread_running(self) -> bool:
         """
         Indicates if there is a thread running with the name self.THREAD_NAME
```
GSVarsha: Please change the docstring here, since the logic has changed

src/instana/collector/base.py
```diff
@@ -93,6 +96,7 @@ def start(self) -> None:
             logger.debug(
                 f"BaseCollector.start non-fatal: call but thread already running (started: {self.started})"
```
GSVarsha: The phrasing here seems a bit weird. Can we rephrase it for better clarity?

src/instana/agent/host.py
```diff
@@ -457,9 +457,13 @@ def diagnostics(self) -> None:
...
                 f"is_collector_thread_running?: {self.collector.is_reporting_thread_running()}"
             )
-            logger.warning(
-                f"background_report_lock.locked?: {self.collector.background_report_lock.locked()}"
+            # RLock doesn't have a locked() method, so we check by trying to acquire
+            lock_acquired = self.collector.background_report_lock.acquire(
+                blocking=False
             )
+            if lock_acquired:
+                self.collector.background_report_lock.release()
```
pvital: I may not be following correctly, but why are you releasing the lock in a function that mainly prints a diagnostic of the agent?

src/instana/collector/host.py
```diff
@@ -43,19 +43,20 @@ def prepare_and_report_data(self) -> None:
         state machine case.
         """
         try:
-            if self.agent.machine.fsm.current == "wait4init":
+            with self.agent.machine._lock:
```
pvital: The `fsm.TheMachine._lock` denotes it's a "private" attribute of the TheMachine class, and it should not be called in that way.

This call is not wrong, but it's not the best practice. I suggest you create a property on `fsm.TheMachine` that returns the value of `_lock`.

src/instana/fsm.py
```diff
@@ -26,14 +26,16 @@ class TheMachine:
     RETRY_PERIOD = 30
     THREAD_NAME = "Instana Machine"
 
-    warnedPeriodic = False
-
     def __init__(self, agent: "HostAgent") -> None:
         logger.debug("Initializing host agent state machine")
 
+        self._lock = threading.RLock()
```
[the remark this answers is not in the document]
CagriYonca: Done

PR #834, opened 2026-01-12
fix: improve confluent kafka's global variable access

CagriYonca:
- Close the consumer span when the poll request returns None.
- Improve the consumer_token and consumer_span access.

### DIFF
--- a/src/instana/instrumentation/kafka/confluent_kafka_python.py
+++ b/src/instana/instrumentation/kafka/confluent_kafka_python.py
@@ -17,8 +17,12 @@ try:
     from instana.span.span import InstanaSpan
     from instana.util.traceutils import get_tracer_tuple
 
-    consumer_token = None
-    consumer_span = contextvars.ContextVar("confluent_kafka_consumer_span")
+    consumer_token = contextvars.ContextVar(
+        "confluent_kafka_consumer_token", default=None
+    )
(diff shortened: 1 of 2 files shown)


src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -178,24 +182,23 @@ def create_span(
...
     def close_consumer_span(span: "InstanaSpan") -> None:
-        global consumer_token
         if span.is_recording():
             span.end()
             consumer_span.set(None)
-        if consumer_token is not None:
-            context.detach(consumer_token)
-            consumer_token = None
+        token = consumer_token.get(None)
+        if token is not None:
```
GSVarsha:
```suggestion
        if token:
```

src/instana/instrumentation/kafka/confluent_kafka_python.py
```diff
@@ -253,6 +256,10 @@ def trace_kafka_poll(
             res = wrapped(*args, **kwargs)
             if res:
                 create_span("poll", res.topic(), res.headers())
+            else:
+                span = consumer_span.get(None)
+                if span is not None:
```
GSVarsha:
```suggestion
                if span:
```

PR #859, opened 2026-04-01
fix: Add support for custom span filtering

CagriYonca:


src/instana/util/span_utils.py
```diff
@@ -50,6 +57,36 @@ def matches_rule(rule_attributes: List[Any], span_attributes: List[Any]) -> bool
     return True
 
 
+def resolve_nested_key(data: Dict[str, Any], key_parts: List[str]) -> Optional[Any]:
```
pvital: Starting with Python 3.9 (our current lower bound supported runtime version), the usage of the built-in collection types like `list`, `dict`, `set`, and `tuple` can be used directly as generic types in annotations, eliminating the need to import capitalized aliases from the typing module (line 4).

I suggest you take the opportunity and update this code to use the built-in collections.


```suggestion
def resolve_nested_key(data: dict[str, Any], key_parts: list[str]) -> Optional[Any]:
```

pvital (changes requested):
Minor request.

src/instana/util/span_utils.py
CagriYonca: Done! (only for the parts I touched)

src/instana/util/span_utils.py
```diff
@@ -49,6 +56,36 @@ def matches_rule(rule_attributes: List[Any], span_attributes: List[Any]) -> bool
     return True
 
 
+def resolve_nested_key(data: dict[str, Any], key_parts: list[str]) -> Optional[Any]:
```
pvital: `Optional[Any]` is redundant — `Any` already includes `None`. The return type is simply `Any`. The import of `Optional` can be dropped entirely.

```suggestion
def resolve_nested_key(data: dict[str, Any], key_parts: list[str]) -> Any:
```

src/instana/util/span_utils.py
```diff
@@ -49,6 +56,36 @@ def matches_rule(rule_attributes: List[Any], span_attributes: List[Any]) -> bool
...
+
+    # Try the longest prefix first so that keys with embedded dots are matched
+    # before shorter splits (e.g. prefer "sdk.custom" over "sdk").
+    for i in range(len(key_parts), 0, -1):
+        candidate = ".".join(key_parts[:i])
+        if candidate in data:
+            remaining = key_parts[i:]
+            if not remaining:
+                return data[candidate]
+            result = resolve_nested_key(data[candidate], remaining)
```
pvital: I suggest that you refactor this function to use a `while` loop instead of a recursive function - a `while` loop is usually faster with constant memory usage than recursion in Python because Python is optimized for iteration, not recursion.

Every recursive call in Python pushes a new frame onto the call stack — local variables, argument bindings, return address, etc. For a deeply nested structure (say, 50 key parts), that's 50 frames allocated and deallocated. The iterative version reuses a single frame throughout. This is relatively costly in Python.

pvital (changes requested):
On top of the following requests, I saw your PR introduces two new maintainability issues [reported](https://github.com/instana/python-sensor/pull/859#issuecomment-4235333638) by SonaQube.  Please, fix them.

src/instana/util/span_utils.py
CagriYonca: Done, please review

GSVarsha:
Verified the changes on scenarios 2 and 4 as described [here](https://github.ibm.com/Varsha-Gs/opentelemetry-instana-python/blob/main/README.md#implementation-scenarios).

### What is verified?
When Instana Python SDK is in charge of the tracer and `requests` instrumentation from OTel instrumentation is also active, we don't see internal calls to the agent such as `POST http://localhost:42699/com.instana.plugin.python.87417`recorded! 🎉

pvital (approved):
It looks good to me.

PR #861, opened 2026-04-07
chore: upgrade Ruff configuration and reformat codebase

pvital:
This PR modernizes the repository’s Ruff setup and applies the resulting formatting and lint cleanup across the codebase and tests.

- Upgraded pre-commit Ruff integration in [`.pre-commit-config.yaml`](.pre-commit-config.yaml)
  - bumped Ruff from `v0.7.0` to `v0.15.9`
  - switched the lint hook from `ruff` to `ruff-check`
  - extended formatting hook coverage to Python and Markdown files
- Added Ruff configuration to [`pyproject.toml`](pyproject.toml)
  - set Python target version to 3.9
  - configured repository exclusions for non-source directories
  - enabled preview mode and concise output
  - defined lint rule selection for style, import ordering, simplification, string modernizations, and copyright checks
- Applied repository-wide Ruff-driven cleanup across source and test files
  - normalized import ordering
  - reformatted multiline logging, context managers, assertions, and long calls
  - simplified membership checks, such as `key in dict` instead of `key in dict.keys()`
  - removed unused imports in several tests
  - cleaned whitespace and newline issues
- Fixed SonarQube failed conditions on new code
  - Deleted `src/instana/util/gunicorn.py` since the code is duplicated in `running_in_gunicorn()` function from `src/instana/log.py`.
  - Adapted code to import `running_in_gunicorn()` from `src/instana/log.py`.
  - Add tests for `running_in_gunicorn()` to increase the coverage on new code.


src/instana/util/config.py
```diff
@@ -394,9 +394,10 @@ def get_disable_trace_configurations_from_yaml() -> Tuple[List[str], List[str]]:
 
 
 def get_disable_trace_configurations_from_local() -> Tuple[List[str], List[str]]:
-    if "tracing" in config:
-        if tracing_disable_config := config["tracing"].get("disable", None):
-            return parse_span_disabling(tracing_disable_config)
+    if "tracing" in config and (
+        tracing_disable_config := config["tracing"].get("disable", None)
```
GSVarsha:
```suggestion
        tracing_disable_config := config["tracing"].get("disable")
```

src/instana/util/span_utils.py
```diff
@@ -56,18 +55,17 @@ def match_key_filter(span_value: str, rule_value: str, match_type: str) -> bool:
...
+        rule_value == "*"
+        or match_type == "strict"
+        and span_value == rule_value
+        or match_type == "contains"
+        and rule_value in span_value
+        or match_type == "startswith"
+        and span_value.startswith(rule_value)
+        or match_type == "endswith"
+        and span_value.endswith(rule_value)
+    )
```
GSVarsha, line 64: Just a slight adjustment for better readability. 
Also I think we don't need the `bool()` wrapper here, as all those comparison operators already return boolean values.

```suggestion
    return (
        rule_value == "*" or
        (match_type == "strict" and span_value == rule_value) or
        (match_type == "contains" and rule_value in span_value) or
        (match_type == "startswith" and span_value.startswith(rule_value)) or
        (match_type == "endswith" and span_value.endswith(rule_value))
    )
```

tests/apps/fastapi_app/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 import uvicorn
 from ...helpers import testenv
-from instana.log import logger
+from instana.log import logger as logger
```
GSVarsha: Do we need this change?

src/instana/collector/helpers/fargate/docker.py
```diff
@@ -1,15 +1,18 @@
...
 
-""" Module to handle the collection of Docker metrics in AWS Fargate """
+"""Module to handle the collection of Docker metrics in AWS Fargate"""
+
 from __future__ import division
+
 from ....log import logger
-from ..base import BaseHelper
 from ....util import DictionaryOfStan
+from ..base import BaseHelper
```
GSVarsha: Can we change these relative imports to absolute imports?

GSVarsha (changes requested):
Just a few suggestions

src/instana/util/config.py
pvital: Interesting to see how Ruff did not catch this [SIM910](https://docs.astral.sh/ruff/rules/dict-get-with-none-default/). It should be part of commit 08083c5057ab388aadab41b7a1d24e1ffe3786e6.

src/instana/util/span_utils.py
pvital: I'll try to apply your suggestion, but probably this will be re-formatted back to the original by Ruff when running the pre-commit rule. Let's see.

tests/apps/fastapi_app/__init__.py
pvital: I don't think so. Will remove this line

GSVarsha (approved):
Looks good to me!

PR #863, opened 2026-04-13
fix: Add pre-commit to circle ci pipeline

CagriYonca:


pvital (commented):
IMHO, the main goal of `pre-commit` is to run before you commit locally during your development to fix all missed project practices. I don't see any value in doing that in the CI pipeline. 

If you want to validate the project rules for a PR, I suggest you create a GitHub Action to execute Ruff, pointing out the necessary errors (like SonarQube does), and blocking the merging. I started something before [here](https://github.com/instana/python-sensor/pull/711) that you can use as a source.

PR #867, opened 2026-04-22
fix: Added exception re-raise in httpx method overwrites

### LINKED ISSUE #866: [Bug]: httpx `handle_request_with_instana` returns `None` instead of re-raising exception
### Problem Description

Greetings,

I've encountered an issue regarding handling httpx requests with Instana. In [this file](https://github.com/instana/python-sensor/blob/main/src/instana/instrumentation/httpx.py), the `httpx.HTTPTransport.handle_request` and `httpx.AsyncHTTPTransport.handle_async_request` methods are overwritten. When tracking is enabled and an exception (for example self-signed certificate exception when `verify=True` is set) appears during the original `handle_request` method execution, the execution jumps to the `except` block:
```py
try:
    request = args[0]
    _set_request_span_attributes(span, request)
    tracer.inject(span.context, Format.HTTP_HEADERS, request.headers)

    response = wrapped(*args, **kwargs)
    _set_response_span_attributes(span, response)
except Exception as e:
    span.record_exception(e)
else:
    return response
```

As you can observe, the exception is not re-raised neither inside nor after `span.record_exception(e)` is executed. This causes the transport to return `None`, which is outside of that method's expected return values. This then causes a `'NoneType' object has no attribute 'stream'` exception later in the request process.

Would it be possible to update the code so it re-raises the exception if it happens?

### Minimal, Complete, Verifiable, Example

You can use a linter like Pylance to detect that these functions can return `None`, even though it's not expected by the type hints.

Without `instana` imported, the following snippet raises an exception, which means it is working as expected:
```py
>>> import httpx
>>> request = httpx.Request("GET", "http://my-website.local")
>>> transport = httpx.HTTPTransport()
>>> print(transport.handle_request(request))
ConnectError: [Errno 8] nodename nor servname provided, or not known
```

When `instana` is imported, the `httpx` call returns `None` instead of raising an exception:
```py
>>> import instana
>>> import httpx
>>> request = httpx.Request("GET", "http://my-website.local")
>>> transport = httpx.HTTPTransport()
>>> print(transport.handle_request(request))
None
```

### Python Version

Python 3.12

### Python Modules

```shell
Not applicable - this is related only to the implementation
```

### Python Environment

```shell
Not applicable - this is related only to the implementation
```

Rafal-Chrzanowski-IBM:
Fixes #866.

Added exception re-raise after exception has been recorded.
Previously `pylance` showed the following warning:
```
Function with declared return type "Response" must return value on all code paths
  "None" is not assignable to "Response"
```
Currently this warning isn't visible anymore.

### DIFF
--- a/src/instana/instrumentation/httpx.py
+++ b/src/instana/instrumentation/httpx.py
@@ -89,6 +89,7 @@ try:
                 _set_response_span_attributes(span, response)
             except Exception as e:
                 span.record_exception(e)
+                raise
             else:
                 return response
 
@@ -118,6 +119,7 @@ try:
                 _set_response_span_attributes(span, response)
             except Exception as e:
                 span.record_exception(e)
+                raise
             else:
                 return response
 
--- a/src/instana/instrumentation/httpx.py
+++ b/src/instana/instrumentation/httpx.py
@@ -82,10 +82,12 @@ try:
         ) as span:
             try:
                 request = args[0]
-                _set_request_span_attributes(span, request)
+                _set_request_span_attributes(span, request)  # Has its own try-except
                 tracer.inject(span.context, Format.HTTP_HEADERS, request.headers)
             except Exception:
-                logger.exception("httpx handle_request_with_instana pre-request:")
+                logger.exception(
+                    "httpx handle_request_with_instana pre-request:", exc_info=True
+                )
 
             try:
                 response = wrapped(*args, **kwargs)
@@ -93,10 +95,7 @@ try:
                 span.record_exception(e)
                 raise
 
-            try:
-                _set_response_span_attributes(span, response)
-            except Exception:
-                logger.exception("httpx handle_request_with_instana post-request:")
+            _set_response_span_attributes(span, response)  # Has its own try-except
 
             return response
 
@@ -119,10 +118,13 @@ try:
         ) as span:
             try:
                 request = args[0]
-                _set_request_span_attributes(span, request)
+                _set_request_span_attributes(span, request)  # Has its own try-except
                 tracer.inject(span.context, Format.HTTP_HEADERS, request.headers)
             except Exception:
-                logger.exception("httpx handle_request_with_instana pre-request:")
+                logger.exception(
+                    "httpx handle_async_request_with_instana pre-request:",
+                    exc_info=True,
+                )
 
             try:
                 response = await wrapped(*args, **kwargs)
@@ -130,10 +132,7 @@ try:
                 span.record_exception(e)
                 raise
 
-            try:
-                _set_response_span_attributes(span, response)
-            except Exception:
-                logger.exception("httpx handle_request_with_instana post-request:")
+            _set_response_span_attributes(span, response)  # Has its own try-except
 
             return response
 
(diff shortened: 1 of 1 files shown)


src/instana/instrumentation/httpx.py
GSVarsha, line 1: Hi @Rafal-Chrzanowski-IBM, thanks for the PR!
According to our spec, we aren't supposed to raise exceptions that are caused by our instrumentation code. But I understand that we shouldn't suppress the exceptions coming from the underlying client library as well. Could you please re-raise the exception only for the original wrapped call as in the snippet I've shared below?

Please update the code (for both the methods) and verify from your end if it fixes the issue. We can then review the PR again, thanks!

``` python
            try:
                request = args[0]
                _set_request_span_attributes(span, request)  # Has its own try-except
                tracer.inject(span.context, Format.HTTP_HEADERS, request.headers)
            except Exception:
                logger.debug("httpx handle_request_with_instana:", exc_info=True)
            
            try:
                response = wrapped(*args, **kwargs)
            except Exception as e:
                span.record_exception(e)
                raise
            
            _set_response_span_attributes(span, response)  # Has its own try-except
```

Rafal-Chrzanowski-IBM: Hi @GSVarsha, thank you for taking a look so quickly. I updated the code according to your suggestions. Please let me know whether I should make any further changes to the PR. Thank you.

src/instana/instrumentation/httpx.py
```diff
@@ -82,15 +82,22 @@ def handle_request_with_instana(
         ) as span:
             try:
                 request = args[0]
-                _set_request_span_attributes(span, request)
+                _set_request_span_attributes(span, request)  # Has its own try-except
                 tracer.inject(span.context, Format.HTTP_HEADERS, request.headers)
+            except Exception:
+                logger.exception(
+                    "httpx handle_request_with_instana pre-request:", exc_info=True
```
GSVarsha:
```suggestion
                    "httpx handle_request_with_instana:", exc_info=True
```

src/instana/instrumentation/httpx.py
```diff
@@ -111,15 +118,23 @@ async def handle_async_request_with_instana(
         ) as span:
             try:
                 request = args[0]
-                _set_request_span_attributes(span, request)
+                _set_request_span_attributes(span, request)  # Has its own try-except
                 tracer.inject(span.context, Format.HTTP_HEADERS, request.headers)
+            except Exception:
+                logger.exception(
+                    "httpx handle_async_request_with_instana pre-request:",
```
GSVarsha:
```suggestion
                    "httpx handle_async_request_with_instana:",
```

src/instana/instrumentation/httpx.py
```diff
@@ -82,15 +82,22 @@ def handle_request_with_instana(
...
-                _set_response_span_attributes(span, response)
             except Exception as e:
                 span.record_exception(e)
-            else:
-                return response
+                raise
+
+            _set_response_span_attributes(span, response)  # Has its own try-except
+
+            return response
```
GSVarsha, line 99:
```suggestion
            _set_response_span_attributes(span, response)  # Has its own try-except
            return response
```

src/instana/instrumentation/httpx.py
```diff
@@ -111,15 +118,23 @@ async def handle_async_request_with_instana(
...
```
GSVarsha, line 135:
```suggestion
            _set_response_span_attributes(span, response)  # Has its own try-except
            return response
```

GSVarsha (changes requested):
Just a few suggestions

GSVarsha (approved):
Looks good to me!

PR #868, opened 2026-04-29
feat: Add adaptive poll rate interval logic

CagriYonca:


src/instana/options.py
```diff
@@ -516,6 +519,32 @@ def set_from(self, res_data: Dict[str, Any]) -> None:
             logger.debug(f"options.set_from: Wrong data type - {type(res_data)}")
             return
 
+        # Extract poll_rate from plugin.python.poll_rate
```
pvital, line 550: Move this whole block to a function, as we have at `set_secrets()` and `set_tracing()`.

src/instana/options.py
pvital, line 564: Extra question: Is this `else` correct?

tests/test_options.py
```diff
@@ -1003,6 +1003,147 @@ def test_set_from_bool(
...
+        """Test setting poll_rate to 5 from announce response"""
+        caplog.set_level(logging.DEBUG, logger="instana")
+        caplog.clear()
+
+        self.standart_options = StandardOptions()
+        test_res_data = {"plugin": {"python": {"poll_rate": 5}}}
+        self.standart_options.set_from(test_res_data)
+
+        assert self.standart_options.poll_rate == 5
+        assert "Poll rate set to 5 seconds from agent configuration" in caplog.messages
```
pvital: These two tests are identical - the only difference is the parameter (poll_rate value) to be tested. 
Please, create a single test function, like `test_set_from_with_valid_poll_rate()`, and use the fixture parametrization from `pytest` to test the values.

tests/test_options.py
```diff
@@ -1003,6 +1003,147 @@ def test_set_from_bool(
...
+        assert self.standart_options.poll_rate == 1
+        assert "Invalid poll_rate value -5, defaulting to 1" in caplog.messages
+
+        # Test with poll_rate = 3
+        caplog.clear()
+        self.standart_options = StandardOptions()
+        test_res_data = {"plugin": {"python": {"poll_rate": 3}}}
+        self.standart_options.set_from(test_res_data)
+        assert self.standart_options.poll_rate == 1
+        assert "Invalid poll_rate value 3, defaulting to 1" in caplog.messages
```
pvital: We have here four identical blocks. Please, write a `test_set_from_with_invalid_poll_rate()` function, and use the fixture parametrization from `pytest` to test the values.

tests/test_options.py
```diff
@@ -1003,6 +1003,147 @@ def test_set_from_bool(
...
+        self.standart_options.set_from(test_res_data)
+        assert self.standart_options.poll_rate == 1
+        assert "Invalid poll_rate type, defaulting to 1" in caplog.messages
+
+        # Test with None
+        caplog.clear()
+        self.standart_options = StandardOptions()
+        test_res_data = {"plugin": {"python": {"poll_rate": None}}}
+        self.standart_options.set_from(test_res_data)
+        assert self.standart_options.poll_rate == 1
```
pvital: Same as [here](https://github.com/instana/python-sensor/pull/868/changes#r3162124718).

pvital (changes requested):
A few requests.

One general question: does your PR impact both `metrics` and `traces` reporting?

src/instana/options.py
CagriYonca: This change is coming from an old ticket, it was mentioned as `Only if the "tracing" does not exist in the response, fall back to the previous way`. I'm sending you the related ticket.

src/instana/options.py
CagriYonca: Done

CagriYonca:
> A few requests.
> 
> One general question: does your PR impact both `metrics` and `traces` reporting?

Now it impacts only the metrics

src/instana/options.py
pvital: I got it. That extra empty line before the `else` confused me. Please, update the code to:

```python
        if "tracing" in res_data:
            self.set_tracing(res_data["tracing"])
        else:
            # Rely on extra headers if no tracing configuration comes from the agent
            if "extraHeaders" in res_data:
                self.set_extra_headers(res_data["extraHeaders"])
```

tests/test_options.py
CagriYonca: Done

tests/test_options.py
CagriYonca: Done

tests/test_options.py
CagriYonca: Done

src/instana/options.py
CagriYonca: Added a descriptive hint to that line.

pvital: Please remove the extra empty line before the `else`. `if-else` blocks should not have empty lines to avoid code reading mistakes.

pvital (changes requested):
One little request you missed.

pvital (approved):
It looks good to me.

PR #864, opened 2026-04-14
ci: Add Ruff linter as GitHub Action.

CagriYonca:
Reopening [an old PR](https://github.com/instana/python-sensor/pull/711) closed some time ago.

@pvital - Executes Ruff as a linter tool to check code patterns.
@CagriYonca - Adds information about pre-commit to the CONTRIBUTING.md

Ruff linter has been tested [here](https://github.com/instana/python-sensor/actions/runs/24397262909).

PR #870, opened 2026-05-04
chore: typing fix for span utils

CagriYonca:
[The ruff action PR](https://github.com/instana/python-sensor/pull/864) requires this fix to properly run the ruff action workflow


pvital (approved):
It looks good to me.

PR #877, opened 2026-05-19
ci: Add 5-days grace period

CagriYonca:
Modified CircleCI:
- Added 5-day grace period:
  - If a new version of a package is published, we'll wait 5 days to avoid zero-day attacks. `pin_safe_versions.py` file will downgrade any installed packages that were released within the 5-day grace period to their latest safe version.
- The latest version of Cassandra wasn't compatible with Python 3.9, so the CircleCI workflow has also been updated to use Python 3.12.
- Each successful run will create artifacts as shown below and these artifacts will be used in the currency report generation.
- Each workflow may have a different version of a package (due to version bounding and cross-dependencies), so we'll look for each and get the highest version among them.
 
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,12 @@ commands:
           command: |
             . venv/bin/activate
             pip install -r <<parameters.requirements>>
+      - run:
+          name: Apply grace period to installed packages
+          command: |
+            . venv/bin/activate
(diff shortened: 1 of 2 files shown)


CagriYonca:
Regarding failing workflows as of today

image

pvital (approved):
It looks good to me.

PR #882, opened 2026-06-18
feat: Add elasticsearch instrumentation

CagriYonca:
- Added elasticsearch instrumentation.
image

### DIFF
--- a/src/instana/instrumentation/elasticsearch.py
+++ b/src/instana/instrumentation/elasticsearch.py
@@ -1,0 +1,857 @@
+# (c) Copyright IBM Corp. 2026
+
+"""
+Elasticsearch instrumentation
+Supports both sync and async clients for elasticsearch
+"""
+
+try:
+    import json
+    import re
+    import time
+    from collections import defaultdict
+    from typing import TYPE_CHECKING, Any, Callable, Coroutine, Optional, Union
+
+    if TYPE_CHECKING:
+        from instana.span.span import InstanaSpan
+        from elasticsearch import AsyncElasticsearch, Elasticsearch
+        from elastic_transport import ObjectApiResponse
+
+    import elasticsearch  # noqa: F401
+    import wrapt
+    from opentelemetry.context import get_current
+    from instana.log import logger
+    from instana.util.traceutils import get_tracer_tuple
+
+    ELASTICSEARCH_INDEX_ATTRIBUTE = "elasticsearch.index"
+    ELASTICSEARCH_ID_ATTRIBUTE = "elasticsearch.id"
+    ELASTICSEARCH_HITS_ATTRIBUTE = "elasticsearch.hits"
+    ELASTICSEARCH_ERROR_ATTRIBUTE = "elasticsearch.error"
+
+    # Regex patterns for URL parsing
+    DOCUMENT_ID_PATTERN = re.compile(r"^/[^/]+/_doc/([^/?]+)")
+    INDEX_PATTERN = re.compile(r"^/([^/?]+)")
+
+    # Map URL _keyword segments to action names (for GET/HEAD/DELETE/other methods)
+    _URL_KEYWORD_ACTION: dict[str, str] = {
+        "_msearch": "msearch",
+        "_mget": "mget",
+        "_bulk": "bulk",
+        "_search": "search",
+        "_update": "update",
+        "_mapping": "indices.getMapping",
+        "_settings": "indices.getSettings",
+    }
+    # Keywords whose action depends on the HTTP method
+    _URL_KEYWORD_METHOD_ACTION: dict[str, dict[str, str]] = {
+        "_doc": {"POST": "index", "PUT": "index", "GET": "get", "DELETE": "delete"},
+        "_create": {"POST": "index", "PUT": "index"},
+        "_mapping": {"PUT": "indices.putMapping"},
+        "_settings": {"PUT": "indices.putSettings"},
+    }
+
+    # Connection cache to avoid repeated URL parsing and store cluster info
+    # Structure: {connection_id: {host, port, cluster_name, last_updated}}
+    _connection_cache: defaultdict[str, dict[str, Any]] = defaultdict(dict)
+
+    # Cluster name cache TTL (5 minutes)
+    CLUSTER_NAME_CACHE_TTL = 300
+
+    def get_connection_id(
+        instance: "Union[Elasticsearch, AsyncElasticsearch]",
+    ) -> Optional[str]:
+        """
+        Generate a unique connection ID for caching.
+        Uses host:port as identifier via elastic-transport node_pool.
+        """
+        try:
+            if hasattr(instance, "transport"):
+                transport = instance.transport
+                if hasattr(transport, "node_pool"):
+                    nodes = list(transport.node_pool.all())
+                    if nodes:
+                        cfg = nodes[0].config
+                        return f"{cfg.host}:{cfg.port}"
+        except Exception:
+            logger.debug("get_connection_id error:", exc_info=True)
+        return None
+
+    def _get_cached_cluster_name(connection_id: str) -> Optional[str]:
+        """Return cached cluster name if still within TTL, otherwise None."""
+        cached = _connection_cache[connection_id]
+        cluster_name = cached.get("cluster_name")
+        if (
+            cluster_name
+            and (time.time() - cached.get("last_updated", 0)) < CLUSTER_NAME_CACHE_TTL
+        ):
+            return cluster_name
+        return None
+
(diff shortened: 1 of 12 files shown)


src/instana/span/registered_span.py
```diff
@@ -263,6 +263,71 @@ def _populate_exit_span_data(self, span: "InstanaSpan") -> None:
             )
             self.data["dynamodb"]["table"] = span.attributes.pop("dynamodb.table", None)
 
+        elif span.name == "elasticsearch":
```
GSVarsha, line 129: `_populate_exit_span_data()` is getting bigger in size. Can we have separate methods like we have for `http`-> `self._collect_http_attributes(span)` for each span type?

CagriYonca: You're right, it's been modified

src/instana/span/registered_span.py
pvital, line 1: I really liked and appreciated this **refactor** by moving each block to its own function - the code gets cleaner to read and understand. :-D

However, since this is a (simple) code **refactor** in a "critical" part of the tracer, I think it's better to separate it into a single commit. Then, if necessary, we can easily revert it.

You can keep the `_collect_elasticsearch_attributes(span)` as part of the ElasticSearch commit, but please move the rest to a `refactor` commit.

src/instana/instrumentation/elasticsearch.py
```diff
@@ -0,0 +1,804 @@
...
+    from instana.log import logger
+    from instana.util.traceutils import get_tracer_tuple
+
+    # Regex patterns for URL parsing
+    DOCUMENT_ID_PATTERN = re.compile(r"^/[^/]+/_doc/([^/?]+)")
+    INDEX_PATTERN = re.compile(r"^/([^/?]+)")
+
+    # Connection cache to avoid repeated URL parsing and store cluster info
+    # Structure: {connection_id: {host, port, cluster_name, last_updated}}
+    _connection_cache: dict[str, dict[str, Any]] = {}
```
pvital: Suggestion. You can use [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) to create `_connection_cache`, then you can better manage the usage of the dictionary. But remember, if asking for a missing key `_connection_cache`, it will return an empty dictionary. WDYT?


```suggestion
    _connection_cache: dict[str, dict[str, Any]] = defaultdict(dict)
```

src/instana/instrumentation/elasticsearch.py
```diff
@@ -0,0 +1,804 @@
...
+        """Return cached cluster name if still within TTL, otherwise None."""
+        if connection_id in _connection_cache:
+            cached = _connection_cache[connection_id]
+            cluster_name = cached.get("cluster_name")
+            if (
+                cluster_name
+                and (time.time() - cached.get("last_updated", 0))
+                < CLUSTER_NAME_CACHE_TTL
+            ):
+                return cluster_name
```
pvital: By using `defaultdict` as suggested before, you can reduce this code to:
```suggestion
        if cached := _connection_cache[connection_id]:
            if (cluster_name := cached.get("cluster_name")) and (time.time() - cached.get("last_updated", 0) < CLUSTER_NAME_CACHE_TTL:
                return cluster_name
```

src/instana/instrumentation/elasticsearch.py
```diff
@@ -0,0 +1,804 @@
...
+        return None
+
+    def _store_cluster_name(connection_id: str, cluster_name: str) -> None:
+        """Persist a discovered cluster name into the connection cache."""
+        if connection_id not in _connection_cache:
+            _connection_cache[connection_id] = {}
+        _connection_cache[connection_id]["cluster_name"] = cluster_name
+        _connection_cache[connection_id]["last_updated"] = time.time()
```
pvital: By using `defaultdict` as suggested before, you can remove this `if` statement:
```suggestion
        _connection_cache[connection_id]["cluster_name"] = cluster_name
        _connection_cache[connection_id]["last_updated"] = time.time()
```

src/instana/instrumentation/elasticsearch.py
```diff
@@ -0,0 +1,804 @@
...
+
+    def discover_cluster_name(instance: Any, connection_id: str) -> Optional[str]:
+        """
+        Discover Elasticsearch cluster name by calling cluster info API (sync).
+        Caches result with TTL to avoid repeated API calls.
+        """
+        try:
+            cached = _get_cached_cluster_name(connection_id)
+            if cached:
+                return cached
```
pvital:
```suggestion
            if cached := _get_cached_cluster_name(connection_id):
                return cached
```

src/instana/instrumentation/elasticsearch.py
```diff
@@ -0,0 +1,804 @@
...
+
+            # perform_request is already instrumented; the span_name == "elasticsearch"
+            # guard inside it prevents recursive tracing of this info() call.
+            if hasattr(instance, "info"):
+                try:
+                    cluster_name = _extract_cluster_name_from_response(instance.info())
+                    if cluster_name:
```
pvital:
```suggestion
                    if cluster_name := _extract_cluster_name_from_response(instance.info())
```

src/instana/instrumentation/elasticsearch.py
```diff
@@ -0,0 +1,804 @@
...
+                return "msearch"
+            if "_mget" in url_lower:
+                return "mget"
+            if "_bulk" in url_lower:
+                return "bulk"
+
+            # Search operations
+            if "_search" in url_lower:
+                return "search"
+
```
pvital: SonarCloud reported a Cognitive Complexity in this function due to the many `if` statements. Why not use something like: 


```suggestion
            url_lower = url.lower()
            if url_lower.startswith("_"):
                 return url_lower[1,]
```

src/instana/instrumentation/elasticsearch.py
```diff
@@ -0,0 +1,804 @@
...
+                # so ES internal header processing (mimetype compat) still works
+                response = wrapped(*args, **kwargs)
+
+                # Extract response metadata
+                extract_response_metadata(span, response)
+
+                # Response status code
+                if hasattr(response, "meta") and hasattr(response.meta, "status"):
+                    status_code = response.meta.status
+                    span.set_attribute(SpanAttributes.HTTP_STATUS_CODE, status_code)
```
pvital: Why do we set the HTTP status code attribute since this is a DB-type operation?

src/instana/span/registered_span.py
CagriYonca: Makes sense, updated!

src/instana/instrumentation/elasticsearch.py
CagriYonca: Doesn't the walrus operator make it harder to read?

src/instana/instrumentation/elasticsearch.py
CagriYonca: Sure, done

src/instana/instrumentation/elasticsearch.py
CagriYonca: You're right, it uses urllib3 to make DB calls but semantically DB span shouldn't have HTTP_STATUS_CODE. It's been removed

src/instana/instrumentation/elasticsearch.py
CagriYonca: Done

src/instana/instrumentation/elasticsearch.py
CagriYonca: Discussed in the meeting, and updated the code

src/instana/instrumentation/elasticsearch.py
GSVarsha: I think this is still not resolved

src/instana/instrumentation/elasticsearch.py
GSVarsha: I think this is still not resolved

src/instana/instrumentation/elasticsearch.py
pvital, line 1: Please, solve the 9 new issues listed in the Sonar report.

pvital (changes requested):

src/instana/instrumentation/elasticsearch.py
CagriYonca: Resolved now

src/instana/instrumentation/elasticsearch.py
CagriYonca: Resolved now

src/instana/instrumentation/elasticsearch.py
CagriYonca: Resolved

pvital (approved):
It looks good to me.

PR #883, opened 2026-06-19
refactor: Remove tekton folder from the codebase

CagriYonca:
We had two similar CI pipelines to run unittests:
- One for daily unittest checks on CircleCI
- Another one for currency checks on Tekton

Now we're beginning to use the result of CircleCI pipeline to run the currency report generation. Tekton-related pipeline codes will be moved into the internal repository.

CI pipeline on Tekton will be removed, the pipeline in the internal repository will pull the artifacts of CircleCI pipeline and generate the reports from them.


GSVarsha (approved):
LGTM
If possible, please update the PR description with a short note on why this is being removed. eg: Moving completely to CircleCI for nightly CI and currency pipelines

PR #884, opened 2026-06-19
fix: Add asyncio task context propagation to env vars

CagriYonca:
Some of LLM technologies, such as Open WebUI and LiteLLM, use asyncio internally, so when we don't set asyncio task context propagation to True manually (which is False by default), internal async threads are created with individual Trace IDs. As a result, you were seeing `SOURCE: Not monitored by Instana` on the Instana UI, instead of appearing as a child of the Open WebUI request.

### DIFF
--- a/src/instana/options.py
+++ b/src/instana/options.py
@@ -16,7 +16,7 @@ BaseOptions - base class for all environments.  Holds settings common to all.
 
 import logging
 import os
-from typing import Any, Dict, Sequence, Tuple
+from typing import Any, Sequence, Union
 
 from instana.configurator import config
 from instana.log import logger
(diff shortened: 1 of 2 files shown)


src/instana/options.py
```diff
@@ -436,6 +417,32 @@ def set_tracing(self, tracing: Dict[str, Any]) -> None:
         # Handle stack trace configuration from agent config
         self.set_stack_trace_from_agent(tracing)
 
+    def _apply_agent_filter_config(self, filter_config: Any) -> None:
```
pvital: I don't like the indiscriminate use of the `Any` type. It gives the impression that you don't want to specify the correct type, and I don't see the need to use type annotations if you're using `Any`.

The `filter_config` argument is used only to call the `parse_filter_rules()` function. This function accepts a `dict[str, Any]` (again, ~Any` here). So I'd suggest following what the `parse_filter_rules()` function accepts, at least.

```suggestion
    def _apply_agent_filter_config(self, filter_config: dict[str, Any]) -> None:
```

pvital (changes requested):

src/instana/options.py
CagriYonca: Done, you're right

pvital (approved):
It looks good to me.

PR #885, opened 2026-06-23
fix: Modify span filtering configuration setting

CagriYonca:
When the agent configuration was used to set filters, it was overwriting the internal filters. With this fix, it'll extend the span-filter rules.


pvital (approved):
It looks good to me.

PR #886, opened 2026-06-24
ci: Add pip-audit control logic

CagriYonca:
Some of the packages, like boto3, have frequent version releases. In our previous approach, we were resetting grace-period counter every time a new release occurred in the grace period. With this new modification, we save the dates of the version releases and after 5 days, we check vulnerabilities of the version with pip-audit package.


pvital (approved):
It looks good to me.

PR #888, opened 2026-07-13
feat: Add twisted instrumentation and unit/integration tests

CagriYonca:
Added twisted webserver instrumentation and related tests.


src/instana/instrumentation/twisted/__init__.py
```diff
@@ -0,0 +1 @@
+# (c) Copyright IBM Corp. 2026
```
pvital: If you don't load or start anything in the `__init__.py` file, it must be empty.

tests/frameworks/test_twisted_client.py
```diff
@@ -0,0 +1,516 @@
...
+        test_span = get_first_span_by_name(spans, "sdk")
+
+        assert server_span
+        assert client_span
+        assert test_span
+
+        assert "X-Capture-This-Too" in server_span.data["http"]["header"]
+        assert server_span.data["http"]["header"]["X-Capture-This-Too"] == "this too"
+        assert "X-Capture-That-Too" in server_span.data["http"]["header"]
+        assert server_span.data["http"]["header"]["X-Capture-That-Too"] == "that too"
```
GSVarsha: The first 8 test methods in `TestTwistedClient` are functionally identical to the 8 tests of the same name in `TestTwistedServer`. Every one of them:

- Looks up `get_first_span_by_name(spans, "urllib3")` — the variable is even named `client_span`, but it holds the `urllib3` span, not a `twisted-client` span
- Never checks for a span named "**twisted-client**" at all

src/instana/instrumentation/twisted/client.py
```diff
@@ -0,0 +1,133 @@
...
+                    parts[1],
+                    agent.options.secrets_matcher,
+                    agent.options.secrets_list,
+                )
+                span.set_attribute("http.params", cleaned_qp)
+
+            span.set_attribute(SpanAttributes.HTTP_METHOD, method_str)
+
+            # Build / augment headers with trace correlation
+            from twisted.web.http_headers import Headers as TwistedHeaders
```
GSVarsha: Please move this to the top, in the outermost `try` block

src/instana/instrumentation/twisted/client.py
```diff
@@ -0,0 +1,133 @@
...
+
+            return deferred
+        except Exception:
+            logger.debug("twisted request_with_instana", exc_info=True)
+            return wrapped(*argv, **kwargs)
+
+    def _finish_tracing(result: Any, span: "InstanaSpan") -> Any:
+        """Callback/errback attached to the Agent.request Deferred."""
+        try:
+            from twisted.python.failure import Failure
```
GSVarsha: Same as [here](https://github.com/instana/python-sensor/pull/888/changes#r3619753709)

src/instana/instrumentation/twisted/client.py
```diff
@@ -0,0 +1,133 @@
...
+
+            # Build / augment headers with trace correlation
+            from twisted.web.http_headers import Headers as TwistedHeaders
+
+            if headers is None or not isinstance(headers, TwistedHeaders):
+                headers = TwistedHeaders({})
+
+            # Capture outgoing request headers
+            headers_dict = {
+                k.decode("latin-1"): v[0].decode("latin-1")
```
GSVarsha: As mentioned in the official twisted doc [here](https://docs.twisted.org/en/stable/api/twisted.web.http_headers.Headers.html), header names are encoded using `ISO-8859-1` (latin-1) and header values are encoded using `UTF-8`.

`getAllRawHeaders()` returns raw bytes in both cases, so the caller must decode each correctly:
```suggestion
                k.decode("latin-1"): v[0].decode("utf-8")
```

src/instana/instrumentation/twisted/client.py
```diff
@@ -0,0 +1,133 @@
...
+            response = result
+            status_code = response.code
+            span.set_attribute(SpanAttributes.HTTP_STATUS_CODE, status_code)
+
+            if status_code >= 400:
+                span.mark_as_errored()
+
+            # Capture response headers
+            headers_dict = {
+                k.decode("latin-1"): v[0].decode("latin-1")
```
GSVarsha: Same as [here](https://github.com/instana/python-sensor/pull/888/changes#r3619803101)

src/instana/instrumentation/twisted/server.py
```diff
@@ -0,0 +1,130 @@
...
+    ) -> Optional[bytes]:
+        try:
+            request: "Request" = argv[0]
+            tracer = get_tracer()
+
+            # Extract parent context from incoming request headers
+            parent_context = None
+            if request.requestHeaders:
+                headers_dict: dict[str, str] = {
+                    k.decode("latin-1"): v[0].decode("latin-1")
```
GSVarsha: Same as [here](https://github.com/instana/python-sensor/pull/888/changes#r3619803101)

src/instana/instrumentation/twisted/server.py
```diff
@@ -0,0 +1,130 @@
...
+                    if isinstance(status_code, int):
+                        _span.set_attribute(
+                            SpanAttributes.HTTP_STATUS_CODE, status_code
+                        )
+                        if status_code >= 500:
+                            _span.mark_as_errored()
+
+                    # Capture response headers
+                    response_hdrs = {
+                        k.decode("latin-1"): v[0].decode("latin-1")
```
GSVarsha: Same as [here](https://github.com/instana/python-sensor/pull/888/changes#r3619803101)

GSVarsha (changes requested):
Few requests

src/instana/instrumentation/twisted/client.py
```diff
@@ -0,0 +1,133 @@
...
+
+            if deferred is not None:
+                deferred.addBoth(_finish_tracing, span)
+
+            return deferred
+        except Exception:
+            logger.debug("twisted request_with_instana", exc_info=True)
+            return wrapped(*argv, **kwargs)
+
+    def _finish_tracing(result: Any, span: "InstanaSpan") -> Any:
```
pvital: If you expect `result` to be `twisted.web.iweb.IResponse` as commented on line 108 below, and the return of this method is the `result` itself, why set the types as `Any`? 
```suggestion
    def _finish_tracing(result: twisted.web.iweb.IResponse, span: "InstanaSpan") -> twisted.web.iweb.IResponse:
```

src/instana/instrumentation/twisted/client.py
```diff
@@ -0,0 +1,133 @@
...
+        """Callback/errback attached to the Agent.request Deferred."""
+        try:
+            from twisted.python.failure import Failure
+
+            if isinstance(result, Failure):
+                span.record_exception(result.value)
+                if span.is_recording():
+                    span.end()
+                return result
```
pvital: When `result` is a `Failure`, the code calls `span.end()` and then returns `result` — bypassing the finally block entirely. That is correct for the happy path. However, if `span.record_exception()` or `span.end()` raises an unexpected exception, control falls through to the `except Exception` block at line 123, and then the `finally` block at line 125 unconditionally calls `span.end()` again (because `span.is_recording()` might still return `True` if `end() `failed before completing). This causes `span.end()` to be called twice.


```suggestion
            if isinstance(result, Failure):
                span.record_exception(result.value)
                return result
```

pvital (changes requested):

src/instana/instrumentation/twisted/__init__.py
GSVarsha: Ack, I don't think we need to copyright an empty file.

src/instana/instrumentation/twisted/client.py
GSVarsha: Since `result` can be either a valid response or a failure, I think a union of them would be more precise.

```suggestion
    def _finish_tracing(result: Union[IResponse, Failure], span: "InstanaSpan") -> Union[IResponse, Failure]:
```

src/instana/instrumentation/twisted/client.py
GSVarsha: Woah that's cool, I just learned that the `finally` block gets executed even with a `return` in the `try` block!
Ack on this one too 👍🏼 

P.S. I found this blog interesting: https://www.raravind.com/blog/data-science/do-you-really-understand-try-finally-in-python

src/instana/instrumentation/twisted/server.py
```diff
@@ -0,0 +1,130 @@
...
+                }
+                parent_context = tracer.extract(Format.HTTP_HEADERS, headers_dict)
+
+            span = tracer.start_span("twisted-server", context=parent_context)
+
+            # Extract the URL components
+            raw_host = request.getHeader(b"host") or b""
+            host = (
+                raw_host.decode("latin-1") if isinstance(raw_host, bytes) else raw_host
+            )
```
pvital: `request.getHeader()` expects an [AnyStr](https://docs.python.org/3/library/typing.html#typing.AnyStr), returning the same or `None`. As an `AnyStr` can be `bytes`  or `str`, I think it would be best if we call it using `str` to already receive it as a decoded string.

```suggestion
            host = request.getHeader("host") or ""
```

src/instana/instrumentation/twisted/server.py
```diff
@@ -0,0 +1,130 @@
...
+            scheme = "https" if request.isSecure() else "http"
+            raw_path = request.path
+            path = (
+                raw_path.decode("latin-1") if isinstance(raw_path, bytes) else raw_path
+            )
```
pvital, line 84: Doesn't the following work?
```suggestion
            path = request.getHeader("path") or ""
```

pvital (changes requested):

src/instana/instrumentation/twisted/server.py
CagriYonca: `request.getHeader("path")` returns None because the path isn't included in the request headers. `request.path` parses the request line to get the path.

tests/frameworks/test_twisted_client.py
CagriYonca: I've fixed the tests and modified them to use `twisted.web.client.Agent`.

src/instana/instrumentation/twisted/client.py
CagriYonca: Done

src/instana/instrumentation/twisted/client.py
CagriYonca: Done

src/instana/instrumentation/twisted/client.py
CagriYonca: Nice catch!

src/instana/instrumentation/twisted/client.py
CagriYonca: Done

src/instana/instrumentation/twisted/server.py
CagriYonca: Done

src/instana/instrumentation/twisted/server.py
CagriYonca: Done

src/instana/instrumentation/twisted/__init__.py
CagriYonca: Done

src/instana/instrumentation/twisted/client.py
CagriYonca: Updated with `Union[IResponse, Failure]`

src/instana/instrumentation/twisted/client.py
CagriYonca: Nice catch, Paulo. Great to know! I've modified this part to ensure the span.end() function runs only once.

src/instana/instrumentation/twisted/server.py
CagriYonca: Done

src/instana/instrumentation/twisted/client.py
```diff
@@ -0,0 +1,130 @@
...
+        try:
+            if isinstance(result, Failure):
+                span.record_exception(result.value)
+            else:
+                response = result
+                status_code = response.code
+                span.set_attribute(SpanAttributes.HTTP_STATUS_CODE, status_code)
+
+                if status_code >= 400:
+                    span.mark_as_errored()
```
pvital: We mark spans as erroneous only if the status code is 500 or higher. Also, we need to provide the reason (or a similar one).

In addition, move this block to be placed after the next block of code (the headers capture).

```suggestion
                if status_code >= 500:
                    span.mark_as_errored({"http.error": response.reason})
```

src/instana/instrumentation/twisted/server.py
pvital: Let's keep it as is, then.

src/instana/instrumentation/twisted/server.py
```diff
@@ -0,0 +1,134 @@
...
+                    }
+                    extract_custom_headers(_span, response_hdrs)
+
+                    if _span.is_recording():
+                        _span.end()
+                except Exception:
+                    logger.debug("twisted finish_with_instana", exc_info=True)
+                finally:
+                    context.detach(request._instana_token)
+                    original_finish()
```
pvital: Couldn't this be placed outside of this inner try block, like the client implementation?

tests/apps/twisted_server/__init__.py
```diff
@@ -0,0 +1,30 @@
+# (c) Copyright IBM Corp. 2026
+
+import os
+import socket
+
+from ...helpers import testenv
+from ..utils import launch_background_thread
```
pvital:
```suggestion
from tests.helpers import testenv
from tests.apps.utils import launch_background_thread
```

tests/frameworks/test_twisted_client.py
```diff
@@ -0,0 +1,460 @@
...
+        assert client_span.p == test_span.s
+        assert server_span.p == client_span.s
+
+        # Error logging
+        assert not test_span.ec
+        assert not client_span.ec
+        assert not server_span.ec
+
+        assert server_span.data["http"]["status"] == 200
+        assert testenv["twisted_server"] + "/" == server_span.data["http"]["url"]
```
pvital: We have the pattern to place the expected value as the second argument of the assertion test.
Please make sure the rest of the code (also the server one) follows the same pattern.

```suggestion
        assert server_span.data["http"]["url"] == testenv["twisted_server"] + "/"
```

tests/frameworks/test_twisted_client.py
```diff
@@ -0,0 +1,460 @@
...
+        assert not server_span.data["http"].get("params")
+        assert server_span.data["http"]["method"] == "GET"
```
pvital: I am not interested in testing if all `server_span` attributes are correct, since we are testing the client side here. Don't you think?

Why not keep it simple and follow this pattern below?

1. Same trace ID (on all three collected spans)
2. Parent relationships (on all three collected spans)
3. Error logging (on all three collected spans)
4. Check span names (on all three collected spans)
5. Check attributes of `client_span` only:
   1. `client_span.data["http"]["status"]`
   2. `client_span.data["http"]["method"]`
   3. `client_span.data["http"]["path"]`
   4. `client_span.data["http"]["url"]`
   5. `client_span.stack`


```suggestion
       # span names
        assert server_span.n == "twisted-server"
        assert client_span.n == "twisted-client"
        assert test_span.data["sdk"]["name"] == "test"
```

pvital (changes requested):
A few more requests.

tests/apps/twisted_server/__init__.py
CagriYonca: Done

tests/frameworks/test_twisted_client.py
CagriYonca: Done

tests/frameworks/test_twisted_client.py
CagriYonca: Done

src/instana/instrumentation/twisted/client.py
CagriYonca: Done

src/instana/instrumentation/twisted/server.py
CagriYonca: Client calls `finish_tracing` with [the Deferred mechanism](https://github.com/twisted/twisted/blob/e4bee3138a21057f3e46c657f463c4cd8ce674f6/src/twisted/internet/defer.py#L668), so it can be called with an argument. Conversely, [server code calls](https://github.com/twisted/twisted/blob/e4bee3138a21057f3e46c657f463c4cd8ce674f6/src/twisted/web/server.py#L270) `finish` function without arguments, so we always need a nested function for that. 

The block has been moved out of the method.

pvital (approved):
It looks good to me.

GSVarsha (approved):
Looks good to me

PR #890, opened 2026-08-07
feat(runtime): Update pollrate behavior and add tests for it

CagriYonca:
## Why

The tracer previously capped `poll_rate` to a maximum of `5` seconds,
silently resetting any higher value back to `1`. This prevented customers
from benefiting from coarser poll rates (e.g. `60 s`, `120 s`) to reduce
observability cost.

Additionally, GC metrics were collected using `gc.get_count()` and
`gc.get_threshold()`, which do not reflect actual GC activity. The correct
source is `gc.get_stats()`, which tracks cumulative `collections`,
`collected`, and `uncollectable` counters per generation — consistent with
the [OpenTelemetry CPython semantic conventions](https://opentelemetry.io/docs/specs/semconv/runtime/cpython-metrics/)
(`cpython.gc.collections`, `cpython.gc.collected_objects`,
`cpython.gc.uncollectable_objects`) and the
[OpenTelemetry Python contrib `system_metrics` instrumentation](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/system_metrics/system_metrics.html),
both of which source these metrics from `gc.get_stats()`.

Finally, the host agent timeout detection used a fixed 60-second window,
causing spurious disconnects when `poll_rate` exceeded 30 seconds.

## What

- **`options.py`**: Replaced `MAX_POLL_RATE = 5` with a `VALID_POLL_RATES`
  list `[1, 5, 10, 20, 30, 60, 120, 180, 240, 300, 360, 420, 480, 540, 600]`.
  `set_poll_rate()` now rounds to the nearest valid value instead of
  rejecting anything outside `{1, 5}`.

- **`runtime.py`**: Switched GC collection from `gc.get_count()` /
  `gc.get_threshold()` to `gc.get_stats()`. Reports per-generation deltas
  for `collections`, `collected`, and `uncollectable` (keys
  `gc.collections0-2`, `gc.collected0-2`, `gc.uncollectable0-2`).

- **`host.py`**: Timeout window is now `max(60, poll_rate * 2)` seconds to
  avoid false resets at high poll rates. Added `_send_heartbeat()` — a
  lightweight `HEAD` request sent when no metrics/spans/profiles are
  produced in a cycle, keeping `last_seen` current between polls.

## Testing

Unit tests are included for all changed behaviours in:
`tests/test_options.py`, `tests/collector/helpers/test_collector_runtime.py`,
`tests/collector/test_host_collector.py`, `tests/agent/test_host.py`

### DIFF
--- a/src/instana/agent/host.py
+++ b/src/instana/agent/host.py
@@ -99,13 +99,31 @@ class HostAgent(BaseAgent):
 
     def is_timed_out(self) -> bool:
         """
-        If we haven't heard from the Instana host agent in 60 seconds, this
-        method will return True.
-        @return: Boolean
+        If we haven't heard from the Instana host agent within the timeout
+        window, this method will return True.  The window is the larger of
+        60 seconds or twice the configured poll_rate so that high poll_rate
+        values (e.g. 120 s) do not cause spurious resets.
+
+        Note: We intentionally read the FSM state directly instead of calling
+        can_send(), because can_send() has a fork-detection side-effect that
+        triggers handle_fork() → reset() and would cause a spurious restart
+        loop on frameworks (e.g. Twisted) that spawn threads with different
+        perceived PIDs.  total_seconds() is used instead of .seconds to
+        correctly handle gaps longer than 24 hours.
+
+        Both "wait4init" and "good2go" states are checked: in wait4init last_seen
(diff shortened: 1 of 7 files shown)


pvital (approved):
Great job @CagriYonca!
It looks good to me.

PR #893, opened 2026-08-20
Add `pymssql` instrumentation

pvital:
This PR adds `pymssql` instrumentation for Microsoft SQL Server, along with tests and CI updates to test it.

- Add `src/instana/instrumentation/pymssql.py`: patches `pymssql.connect`(and the `Connect` alias) via the shared `ConnectionFactory` with `module_name="mssql"`
- Extend `CursorWrapper._collect_kvs` in `pep0249.py`: host now falls back from 'host' to 'server' kwarg (`pymssql` uses server=)
- Add "mssql" to `EXIT_SPANS` tuple in `span/kind.py` (alphabetical order)
- Add `_collect_mssql_attributes` method to `RegisteredSpan` and wire it into `_populate_exit_span_data` after the `mysql` branch
- Register `pymssql` module in boot_agent() import block
- Add `tests/clients/test_pymssql.py`: `TestPyMSSQL` class with autouse fixture that skips gracefully when SQL Server is unreachable; covers all three user stories:
  - US1: vanilla query, basic query, root exit span, parametrized DML statement attributes, cursor context manager
  - US2: parametrized error capture (syntax error, missing table), no-error-on-success assertion
  - US3: executemany parametrized across 2-row and 5-row batches
  - SQLAlchemy bypass (no double-span)
- Add `mssql_*` keys to `tests/helpers.py` (MSSQL_HOST/PORT/DATABASE/USER/SA_PASSWORD env vars, defaults to 127.0.0.1:1433)
- Add `pymssql>=2.2.0` to tests/requirements.txt
- Add the `mcr.microsoft.com/azure-sql-edge` container image into `docker-compose.yml` for local testing.
- Add the `mcr.microsoft.com/mssql/server:2022-latest` container image into the CircleCI configuration for testing across all Python versions.

### DIFF
--- a/src/instana/instrumentation/pep0249.py
+++ b/src/instana/instrumentation/pep0249.py
@@ -1,7 +1,7 @@
 # (c) Copyright IBM Corp. 2021
 # (c) Copyright Instana Inc. 2018
 
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, Optional, Union
 
 # This is a wrapper for PEP-0249: Python Database API Specification v2.0
 import wrapt
@@ -24,8 +24,8 @@ class CursorWrapper(wrapt.ObjectProxy):
         self,
         cursor: Any,
         module_name: str,
-        connect_params: Optional[List[Union[str, Dict[str, Any]]]] = None,
-        cursor_params: Optional[Dict[str, Any]] = None,
+        connect_params: Optional[list[Union[str, dict[str, Any]]]] = None,
+        cursor_params: Optional[dict[str, Any]] = None,
     ) -> None:
         super(CursorWrapper, self).__init__(wrapped=cursor)
         self._module_name = module_name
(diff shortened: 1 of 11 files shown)


GSVarsha (approved):
Looks good to me!

src/instana/span/span.py
```diff
@@ -170,6 +170,8 @@ def record_exception(
                 self.set_attribute("lambda.error", message)
             elif self.name.startswith("kafka"):
                 self.set_attribute("kafka.error", message)
+            elif self.name == "mssql":
+                self.set_attribute("mssql.error", message[1])
```
CagriYonca: Are you sure that it should be set to `message[1]`?

src/instana/instrumentation/pep0249.py
```diff
@@ -52,9 +52,15 @@ def _collect_kvs(
                     self._connect_params[1][db_parameter_name],
                 )
```
CagriYonca, line 54: Since you're changing the file, should we refactor `Dict`, `List`, `Tuple` to `dict`, `list`, `tuple`? We can do this for any file we touch.

CagriYonca (changes requested):
Only two questions, good job!

src/instana/span/span.py
pvital: Good catch @CagriYonca. Fixed.

src/instana/instrumentation/pep0249.py
pvital: Modified. Thanks for remembering me of that.

CagriYonca (approved):
Looks good to me!

PR #894, opened 2026-08-24
feat: Add native kafka-python library support.

CagriYonca:


.circleci/config.yml
```diff
@@ -355,7 +358,10 @@ workflows:
       - py39gevent
       - py312aws
       - py312cassandra
-      - py313kafka
+      - pykafka:
+          matrix:
+            parameters:
+              py-version: ["3.9", "3.12", "3.13"]
```
pvital: I would suggest changing 3.13 to the latest stable version, 3.14.
```suggestion
              py-version: ["3.9", "3.12", "3.14"]
```

tests/requirements-kafka.txt
```diff
@@ -1,5 +1,4 @@
 -r requirements-minimal.txt
 mock>=2.0.0
 confluent-kafka>=2.0.0
-kafka-python>=2.0.0; python_version < "3.12"
-kafka-python-ng>=2.0.0; python_version >= "3.12"
+kafka-python>=2.0.3
```
pvital: Let's keep the previous lower-bound version; remove the conditional and the `kafka-python-ng` line.
```suggestion
kafka-python>=2.0.0
```

tests/requirements-pre315.txt
```diff
@@ -45,5 +45,5 @@ urllib3>=1.26.5
 httpx>=0.27.0
 gevent>=23.9.0.post1
 confluent-kafka>=2.0.0
-kafka-python-ng>=2.0.0
+kafka-python>=2.0.3
```
pvital:
```suggestion
kafka-python>=2.0.0
```

.circleci/config.yml
pvital, line 1: I was thinking whether it wouldn't be better to remove the `KAFKA_TEST` environment variable and make all Kafka tests part of the default pipeline. WDYT?

pvital (changes requested):

.circleci/config.yml
CagriYonca: If the Kafka broker initialises a bit late or only behaves flakily because of async calls, the entire pipeline will run again. It's a bit flakier than other images.

tests/requirements-kafka.txt
CagriYonca: Done

tests/requirements-pre315.txt
CagriYonca: Done

.circleci/config.yml
CagriYonca: Done

.circleci/config.yml
pvital: Got it. Let's keep it as an isolated pipeline.

pvital (approved):
It looks good to me.

PR #900, opened 2026-09-01
feat(configuration): Mark 4xx calls as erroneous spans and adapt current instrumentations.

CagriYonca:
feat: HTTP 4xx exit span error classification

Implements opt-in 4xx error classification for HTTP exit spans per the
tracing specification. Entry spans are never affected.

### What changed

**Configuration**
- Two new opt-in settings under `tracing.http.exit`:
  - `classify-all-4xx-as-errors` (boolean) — marks all 4xx responses as errors on exit spans
  - `classify-as-errors` (list of integers, 400–499) — marks only the listed codes as errors
- `classify-as-errors` takes full precedence over `classify-all-4xx-as-errors` when both are set
- Three configuration sources supported (in priority order):
  1. `INSTANA_TRACING_HTTP_EXIT_CLASSIFY_AS_ERRORS=401,403`
  2. `INSTANA_TRACING_HTTP_EXIT_CLASSIFY_ALL_4XX_AS_ERRORS=true`
  3. Agent config / `INSTANA_CONFIG_PATH` YAML
- Invalid env var values are ignored with a warning; unrecognised values for `CLASSIFY_ALL` keep the default (`false`)

**`http.error` format**
- `http.error` attribute now uses `"<code> <reason>"` format (e.g. `"404 Not Found"`) via a new `http_error_message()` helper, consistent across all four clients: `aiohttp`, `httpx`, `tornado`, `urllib3`

### Tests
- `tests/clients/test_4xx_classification.py` — end-to-end classification tests
- `tests/test_options.py` — configuration precedence, invalid value handling, YAML fallback
- `tests/util/test_http_utils.py` — `http_error_message()` and `should_mark_http_exit_as_error()` unit tests

### DIFF
--- a/src/instana/instrumentation/aiohttp/client.py
+++ b/src/instana/instrumentation/aiohttp/client.py
@@ -2,21 +2,22 @@
 # (c) Copyright Instana Inc. 2019
 
 
-from types import SimpleNamespace
-from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, Tuple
-
-import wrapt
-from opentelemetry.semconv.trace import SpanAttributes
-
-from instana.log import logger
-from instana.propagators.format import Format
-from instana.singletons import agent
-from instana.util.secrets import strip_secrets_from_query
-from instana.util.traceutils import extract_custom_headers, get_tracer_tuple
-
 try:
+    from collections.abc import Awaitable, Callable
+    from types import SimpleNamespace
+    from typing import TYPE_CHECKING, Any
+
     import aiohttp
+    import wrapt
     from opentelemetry.context import get_current
+    from opentelemetry.semconv.trace import SpanAttributes
+
+    from instana.log import logger
+    from instana.propagators.format import Format
+    from instana.singletons import agent
+    from instana.util.http import should_mark_http_exit_as_error
+    from instana.util.secrets import strip_secrets_from_query
+    from instana.util.traceutils import extract_custom_headers, get_tracer_tuple
 
     if TYPE_CHECKING:
         from aiohttp.client import ClientSession
@@ -65,8 +66,9 @@ try:
 
                 extract_custom_headers(span, params.response.headers)
 
-                if params.response.status >= 500:
-                    span.mark_as_errored({"http.error": params.response.reason})
+                if should_mark_http_exit_as_error(params.response.status, agent.options):
+                    error_msg = f"{params.response.status} {params.response.reason}"
+                    span.mark_as_errored({"http.error": error_msg})
 
                 if span.is_recording():
                     span.end()
@@ -92,8 +94,8 @@ try:
     def init_with_instana(
         wrapped: Callable[..., Awaitable["ClientSession"]],
         instance: aiohttp.client.ClientSession,
-        args: Tuple[int, str, Tuple[object, ...]],
-        kwargs: Dict[str, Any],
+        args: tuple[int, str, tuple[object, ...]],
+        kwargs: dict[str, Any],
     ) -> object:
         instana_trace_config = aiohttp.TraceConfig()
         instana_trace_config.on_request_start.append(stan_request_start)
--- a/src/instana/instrumentation/tornado/client.py
+++ b/src/instana/instrumentation/tornado/client.py
@@ -4,7 +4,7 @@
 
 try:
     import functools
-    from typing import TYPE_CHECKING, Any, Callable, Dict, Tuple
+    from typing import TYPE_CHECKING, Any, Callable
 
     import tornado
     import wrapt
@@ -23,6 +23,7 @@ try:
     from instana.propagators.format import Format
     from instana.singletons import agent, get_tracer
     from instana.span.span import get_current_span
+    from instana.util.http import should_mark_http_exit_as_error
     from instana.util.secrets import strip_secrets_from_query
     from instana.util.traceutils import extract_custom_headers
 
@@ -30,8 +31,8 @@ try:
     def fetch_with_instana(
         wrapped: Callable[..., object],
         instance: "AsyncHTTPClient",
-        argv: Tuple[object, ...],
-        kwargs: Dict[str, Any],
+        argv: tuple[object, ...],
+        kwargs: dict[str, Any],
     ) -> "Future":
         try:
             parent_span = get_current_span()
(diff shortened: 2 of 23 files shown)


src/instana/options.py
```diff
@@ -120,6 +125,47 @@ def set_trace_configurations(self) -> None:
...
+                    "Ignoring unrecognised value for"
+                    " INSTANA_TRACING_HTTP_EXIT_CLASSIFY_ALL_4XX_AS_ERRORS:"
+                    f" '{raw}', expected true or false"
+                )
+        elif "INSTANA_CONFIG_PATH" in os.environ:
+            classify_all, codes = get_http_exit_classification_from_yaml()
+            if codes:
+                self.http_exit_classify_as_errors = codes
+            elif classify_all:
+                self.http_exit_classify_all_4xx_as_errors = True
```
pvital: Move this to a specific method like the `set_disable_trace_configurations()` and others. 
It's too long and hard to understand that this is related to setting up the list of 4xx status codes to use as error codes. 

It was even difficult to select all those lines to provide this comment.

src/instana/options.py
```diff
@@ -120,6 +125,47 @@ def set_trace_configurations(self) -> None:
...
+                self.http_exit_classify_as_errors = codes
+        elif "INSTANA_TRACING_HTTP_EXIT_CLASSIFY_ALL_4XX_AS_ERRORS" in os.environ:
+            raw = os.environ["INSTANA_TRACING_HTTP_EXIT_CLASSIFY_ALL_4XX_AS_ERRORS"]
+            if raw.lower() in ("true", "false", "1"):
+                self.http_exit_classify_all_4xx_as_errors = is_truthy(raw)
+            else:
+                logger.warning(
+                    "Ignoring unrecognised value for"
+                    " INSTANA_TRACING_HTTP_EXIT_CLASSIFY_ALL_4XX_AS_ERRORS:"
+                    f" '{raw}', expected true or false"
```
pvital: No need to process all of this information. You have already set `http_exit_classify_all_4xx_as_errors = False` as the default value, so let's only accept a "truthy" value here. 


```suggestion
            self.http_exit_classify_all_4xx_as_errors = is_truthy(
                os.environ["INSTANA_TRACING_HTTP_EXIT_CLASSIFY_ALL_4XX_AS_ERRORS"]
            )
            logger.debug(
                "INSTANA_TRACING_HTTP_EXIT_CLASSIFY_ALL_4XX_AS_ERRORS = True"
            )
```

src/instana/options.py
pvital: Let's ensure the `self.http_exit_classify_as_errors` and `self.http_exit_classify_all_4xx_as_errors` are correct:
```suggestion
        elif "INSTANA_CONFIG_PATH" in os.environ:
            classify_all, codes = get_http_exit_classification_from_yaml()
            if codes:
                self.http_exit_classify_as_errors = codes
                self.http_exit_classify_all_4xx_as_errors = False
            elif classify_all:
                self.http_exit_classify_all_4xx_as_errors = True
                self.http_exit_classify_as_errors = []
```
Actually, if you really guarantee the return of get_http_exit_classification_from_yaml() will be always correct, you can make it simpler:
```suggestion
        elif "INSTANA_CONFIG_PATH" in os.environ:
            self.http_exit_classify_all_4xx_as_errors, self.http_exit_classify_as_errors = get_http_exit_classification_from_yaml()
```

src/instana/instrumentation/aiohttp/client.py
```diff
@@ -65,8 +66,8 @@ async def stan_request_end(
 
                 extract_custom_headers(span, params.response.headers)
 
-                if params.response.status >= 500:
-                    span.mark_as_errored({"http.error": params.response.reason})
+                if should_mark_http_exit_as_error(params.response.status, agent.options):
+                    span.mark_as_errored({"http.error": http_error_message(params.response.status)})
```
pvital: I like the idea of having a method to set the expected value with the status code and reason. However, I don't think that's totally reusable since not all frameworks work with the same encoding or return pattern - you can see your Twisted change; it doesn't fit. In addition, users can customize the `reason` depending on the behavior of their systems, so not all messages will provide the correct reason.

I suggest you remove the `http_error_message()` method and use the f-string format for that.
```suggestion
                    error_msg = f"{params.response.status} {params.response.reason}"
                    span.mark_as_errored({"http.error": error_msg)
```

src/instana/instrumentation/tornado/client.py
```diff
@@ -89,8 +89,9 @@ def finish_tracing(future: "Future", span: "InstanaSpan") -> None:
         try:
             response = future.result()
             span.set_attribute(SpanAttributes.HTTP_STATUS_CODE, response.code)
-
             extract_custom_headers(span, response.headers)
+            if should_mark_http_exit_as_error(response.code, agent.options):
+                span.mark_as_errored({"http.error": http_error_message(response.code)})
```
pvital: Like [here](https://github.com/instana/python-sensor/pull/900/changes#r3915744372):
```suggestion
                error_msg = f"{response.code} {response.body}"
                span.mark_as_errored({"http.error": error_msg)
```

src/instana/instrumentation/tornado/server.py
```diff
@@ -77,6 +77,7 @@ def set_default_headers_with_instana(
         span = instance.request._instana
         tracer = get_tracer()
         tracer.inject(span.context, Format.HTTP_HEADERS, instance._headers)
+        return wrapped(*argv, **kwargs)
```
pvital, line 80: Good catch! 👍

pvital (changes requested):

src/instana/instrumentation/tornado/client.py
CagriYonca: Done

src/instana/instrumentation/aiohttp/client.py
CagriYonca: Done

src/instana/options.py
CagriYonca: Done

src/instana/options.py
CagriYonca: Yeah, you're right, I missed taking this part out of the function 👍

pvital (approved):
It looks good to me.

PR #901, opened 2026-09-08
fix(sqlalchemy): use connection span in after cursor execute

CagriYonca:
### What changed

`receive_after_cursor_execute` in `sqlalchemy.py` had two bugs:

1. **Guard check was dead code.** `get_tracer_tuple()` returns a tuple, so `if not tracer:` was always `False` — the early-return path never ran.

2. **Wrong span was closed.** `get_current_span()` reads from the global OTel context, which could be a different span than the one opened in `before_cursor_execute`. This corrupted the context stack and caused subsequent child spans (e.g. Redis) to be silently dropped.

**Fix:** close `conn.span` directly — the span stored on the connection object by `before_cursor_execute` — instead of reading from global context.

### Tests

- `test_if_not_tracing`: added `assert len(spans) == 0`
- `test_context_restored_after_query`: new test asserting the parent span is active again after a SQL query completes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,4 +106,3 @@ uv.lock
 # Sandbox
 sandbox/
 .bob/
-.serena/
--- a/src/instana/instrumentation/sqlalchemy.py
+++ b/src/instana/instrumentation/sqlalchemy.py
@@ -9,7 +9,7 @@ from opentelemetry import context, trace
 from opentelemetry.context import get_current
 
 from instana.log import logger
-from instana.span.span import InstanaSpan, get_current_span
+from instana.span.span import InstanaSpan
 from instana.span_context import SpanContext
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -106,3 +106,4 @@ uv.lock
 # Sandbox
 sandbox/
 .bob/
+.serena/
```
pvital: Please, remove this.

pvital (changes requested):
Minor request

.gitignore
CagriYonca: This is one of Bob's built-in skills and it's also useful. Can we have it in .gitignore?

CagriYonca: It's been removed

pvital (approved):
It looks good to me.

PR #903, opened 2026-09-16
fix: defer boot_agent until after eventlet monkey_patch to prevent recursion error

CagriYonca:
## Problem & Root Cause

When running under Gunicorn with the `eventlet` worker class, Instana's early auto-instrumentation eagerly imports third-party libraries in the master process before `eventlet.monkey_patch` executes.

Modules such as `urllib3` and `pymongo` capture module-level references to the unpatched `ssl.SSLContext`. When Eventlet subsequently patches the `ssl` module, `ssl.SSLContext` is replaced with its green subclass. Initializing SSL connections later causes `super` lookups inside property setters to resolve the global green subclass instead of the original class, resulting in an infinite recursion and a `RecursionError`.

## Solution

This PR defers `boot_agent` execution when `eventlet` is present in the environment:

- Wraps `eventlet.monkey_patch` using `wrapt`.
- Defers `boot_agent` and all eager instrumentation imports until after `eventlet.monkey_patch` completes in the forked worker process.
- Suppresses sensor boot in the Gunicorn arbiter process, ensuring instrumentation runs only in worker processes.

## Changes

- **`src/instana/__init__.py`**: Added `_defer_boot_until_eventlet_patch` to defer agent boot until worker post-patch.
- **`tests/frameworks/test_eventlet_autotrace.py`**: Added unit tests covering deferred boot registration, worker execution, and arbiter suppression.


src/instana/__init__.py
```diff
@@ -59,10 +59,7 @@
 
 
 def load(_: object) -> None:
-    """
-    Method used to activate the Instana sensor via AUTOWRAPT_BOOTSTRAP
-    environment variable.
-    """
+    """Activate the Instana sensor via AUTOWRAPT_BOOTSTRAP environment variable."""
```
pvital: Since we are changing the docstring, let's correct the names. 😅 
```suggestion
    """Activate the Instana Tracer via the AUTOWRAPT_BOOTSTRAP environment variable."""
```

src/instana/__init__.py
```diff
@@ -260,8 +305,19 @@ def _start_profiler() -> None:
...
+        # Eventlet deferred boot: if eventlet is present in the environment,
+        # defer boot_agent() until after eventlet.monkey_patch() has been called
+        # (which happens in each worker process after fork).  This prevents the
+        # ssl.SSLContext global-rebind RecursionError seen with gunicorn eventlet workers.
+        if (
+            (is_autowrapt_instrumented() or is_webhook_instrumented())
+            and "INSTANA_DISABLE_AUTO_INSTR" not in os.environ
+            and importlib_util.find_spec("eventlet")
+        ):
+            _defer_boot_until_eventlet_patch()
```
pvital: Let me understand here. If `gevent` and `eventlet` are used at the same time (I don't know if that's possible or not), first we apply the gevent monkey patch and then the eventlet monkey patch?

Anyway, I see some repetitive code here; I guess we can reduce this to something like (consider the change starting on line 299):
```suggestion
if (
    (is_autowrapt_instrumented() or is_webhook_instrumented())
    and "INSTANA_DISABLE_AUTO_INSTR" not in os.environ
):
    # Automatic gevent monkey patching
    # unless auto instrumentation is off, then the customer should do manual gevent monkey patching
    if importlib_util.find_spec("gevent"):
        apply_gevent_monkey_patch()

    # Eventlet deferred boot: if eventlet is present in the environment,
    # defer boot_agent() until after eventlet.monkey_patch() has been called
    # (which happens in each worker process after fork).  This prevents the
    # ssl.SSLContext global-rebind RecursionError seen with gunicorn eventlet workers.
    if  importlib_util.find_spec("eventlet"):
        _defer_boot_until_eventlet_patch()
```

src/instana/__init__.py
```diff
@@ -260,8 +305,19 @@ def _start_profiler() -> None:
...
+        else:
+            # AutoProfile
+            if "INSTANA_AUTOPROFILE" in os.environ:
+                _start_profiler()
```
pvital: The AutoProfile startup must be out of the if block. 
BTW, let's check the value of the environment variable, accepting only truthy ones:

```suggestion
        # AutoProfile
        if is_truthy(os.environ.get("INSTANA_AUTOPROFILE", None):
            _start_profiler()
```

pvital (changes requested):
Great work @CagriYonca.
I have a few considerations.

src/instana/__init__.py
CagriYonca: I've refactored the block as you suggested, the shared condition is now extracted into a single outer block, with gevent and eventlet checks nested inside.

gevent and eventlet are not used simultaneously, these are Gunicorn's worker classes and they serve the same purpose. The two inner checks are independent branches for each case.

src/instana/__init__.py
CagriYonca: Done

src/instana/__init__.py
```diff
@@ -251,17 +296,24 @@ def _start_profiler() -> None:
...
+            # ssl.SSLContext global-rebind RecursionError seen with gunicorn eventlet workers.
+            if importlib_util.find_spec("eventlet"):
+                _defer_boot_until_eventlet_patch()
 
         # AutoProfile
-        if "INSTANA_AUTOPROFILE" in os.environ:
+        if is_truthy(os.environ.get("INSTANA_AUTOPROFILE", None)):
             _start_profiler()
 
         boot_agent()
```
GSVarsha, line 319: If `is_worker` is `True`, `boot_agent()` is called twice

src/instana/__init__.py
```diff
@@ -100,16 +97,59 @@ def key_to_bool(k: str) -> bool:
...
+        is_worker = False
+        try:
+            with open(f"/proc/{os.getppid()}/cmdline", "rb") as fh:
+                is_worker = b"gunicorn" in fh.read()
+        except OSError:
+            pass
+        if is_worker:
+            if is_truthy(os.environ.get("INSTANA_AUTOPROFILE", None)):
+                _start_profiler()
+            boot_agent()
```
GSVarsha, line 136: Same as [here](https://github.com/instana/python-sensor/pull/903/changes#r4035575267)

src/instana/__init__.py
```diff
@@ -100,16 +97,59 @@ def key_to_bool(k: str) -> bool:
...
-    parses that environment variable or returns the defaults.
+    import wrapt
+
+    def _after_monkey_patch(
+        wrapped: object,
+        instance: object,
+        args: tuple[object, ...],
+        kwargs: dict[str, object],
+    ) -> Optional[object]:
+        result = wrapped(*args, **kwargs)
```
GSVarsha, line 123: `eventlet.monkey_patch()` returns `None`

```suggestion
        wrapped(*args, **kwargs)
```

src/instana/__init__.py
```diff
@@ -100,16 +97,59 @@ def key_to_bool(k: str) -> bool:
...
+        try:
+            with open(f"/proc/{os.getppid()}/cmdline", "rb") as fh:
+                is_worker = b"gunicorn" in fh.read()
+        except OSError:
+            pass
+        if is_worker:
+            if is_truthy(os.environ.get("INSTANA_AUTOPROFILE", None)):
+                _start_profiler()
+            boot_agent()
+        return result
```
GSVarsha, line 137:
```suggestion
```
Reason: same as [here](https://github.com/instana/python-sensor/pull/903/changes#r4035727089)

src/instana/__init__.py
```diff
@@ -100,16 +97,59 @@ def key_to_bool(k: str) -> bool:
...
-    in the LAMBDA_HANDLER environment variable.  This function searches for and
-    parses that environment variable or returns the defaults.
+    import wrapt
+
+    def _after_monkey_patch(
+        wrapped: object,
+        instance: object,
+        args: tuple[object, ...],
+        kwargs: dict[str, object],
+    ) -> Optional[object]:
```
GSVarsha, line 122:
```suggestion
    ) -> None:
```
Reason: same as [here](https://github.com/instana/python-sensor/pull/903/changes#r4035727089)

src/instana/__init__.py
```diff
@@ -14,7 +14,7 @@
 import os
 import sys
 from importlib import util as importlib_util
-from typing import Tuple
+from typing import Optional
```
GSVarsha, line 17:
```suggestion
```
Reason: same as [here](https://github.com/instana/python-sensor/pull/903/changes#r4035727089)

GSVarsha (changes requested):
Few requests.

PR #904, opened 2026-09-17
fix(tracer): support converting OpenTelemetry parent span context

arjun-rajappa:
Allow OpenTelemetry SpanContext instances in start_span by converting valid non-Instana parent contexts into Instana SpanContext objects instead of raising a TypeError. If the context is invalid, fall back to None. Include debug logs for parent context conversion and invalid context resets.


pvital (changes requested):
@arjun-rajappa, can you add specific `start_span` test cases with `context` arguments with OTel and Instana `SpanContext`? You can use the [Pytest parametrization](https://docs.pytest.org/en/stable/example/parametrize.html) for that.

